PR #16, opened 2017-05-30
Add --exclude=<name> argument to exclude plugins or themes from updating

### LINKED ISSUE #14: Add --exclude=<name> argument for excluding plugins and themes from update
If you want to update all plugins and themes except for a specific few, it would be helpful to have `--exclude=<name>` to exclude one or more from update.

Nikschavan:
Fixes #14

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -498,6 +498,9 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 *
 	 * [--all]
 	 * : If set, all plugins that have updates will be updated.
+	 * 
+	 * [--exclude=<name>]
+	 * : Comma separated list of plugin names that should be excluded from updating.
 	 *
 	 * [--minor]
 	 * : Only perform updates for minor releases (e.g. from 1.3 to 1.4 instead of 2.0)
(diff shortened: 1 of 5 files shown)


src/Theme_Command.php
```diff
@@ -589,6 +589,9 @@ public function get( $args, $assoc_args ) {
 	 *
 	 * [--all]
 	 * : If set, all themes that have updates will be updated.
+	 * 
+	 * [--exclude=<name>]
+	 * : Comma separated list of plugin names that should be excluded from updating.
```
danielbachhuber: This should be "theme names"

src/Plugin_Command.php
```diff
@@ -498,6 +498,9 @@ protected function install_from_repo( $slug, $assoc_args ) {
 	 *
```
danielbachhuber: Can you include functional tests for using this argument with both themes and plugins?

Nikschavan: How are the plugin updates emulated in the test environment?

danielbachhuber: > How are the plugin updates emulated in the test environment?

Have you looked at the existing tests? How do you think they could be emulated?

danielbachhuber:
Thanks for your work on this, @Nikschavan

PR #23, opened 2017-06-29
Update README for new `--exclude=<name>` argument

danielbachhuber:
From #16


danielbachhuber:
Related https://github.com/wp-cli/scaffold-package-command/issues/116

PR #31, opened 2017-08-18
Wrap all uses of get_plugins() with the "all_plugins" filter

### LINKED ISSUE #30: Suggestion: use 'all_plugins' filter in all commands that affect all plugins
I noticed that the `wp plugin` cli command doesn't make use of the `all_plugins` filter which is used in the WP_Plugins_List_Table controller. This filter can be used in the dashboard to modify the list of plugins which are visible in the plugins list table, among other things. For example, I'm using it in a project to hide a dev-only plugin from production environments.

This means that the list of plugins available through `wp plugin list` is different from that shown on `wp-admin/plugins.php`. Also, other inconsistencies exist - the biggest I've come across being that `wp plugin activate --all` may activate different plugins than the "Activate" bulk action on the plugins page.   

I think that it would be good for all the `wp plugin` subcommands to wrap any `get_plugins()` calls with this filter. I'm happy to make a PR for it, but wanted to get a second opinion before jumping in with it.

goldenapples:
Uses the same filter used in the core list table at plugins.php, to ensure that plugin operations performed through wp-cli have the same effect as those done through the admin interface.

Includes some functional tests to ensure that if a plugin is hidden by a filter on "all_plugins":
- it will not be displayed in the output of `wp plugin list`
- it will not be activated by `wp plugin activate --all`
- it will not be deactivated by `wp plugin deactivate --all`

Fixes #30

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -426,8 +426,8 @@ Feature: Manage WordPress plugins
     And a wp-content/plugins/handbook/functionality-for-pages.php file:
       """
       <?php
-	  /**
-	   * Plugin Name: Handbook Functionality for Pages
+      /**
+       * Plugin Name: Handbook Functionality for Pages
        * Description: Adds handbook-like table of contents to all Pages for a site. Covers Table of Contents and the "watch this page" widget
        * Author: Nacin
        */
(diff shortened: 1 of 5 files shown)


features/plugin.feature
```diff
@@ -426,8 +426,8 @@ Feature: Manage WordPress plugins
     And a wp-content/plugins/handbook/functionality-for-pages.php file:
       """
       <?php
-	  /**
-	   * Plugin Name: Handbook Functionality for Pages
+      /**
+       * Plugin Name: Handbook Functionality for Pages
```
goldenapples: Not related to this PR, but there were a couple tabs in here I cleaned up.

goldenapples:
This is ready for review.

goldenapples:
Thanks! :tada:

Issue #36, opened 2017-08-29
Improve 'wp plugin' command description


hearvox:
Improve command description in:
https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php
Command: wp plugin
Current description: Manage plugins.

Proposed description:
Manages site plugins, including installs, activations, and updates.

miya0001:
Related:
https://github.com/wp-cli/wp-cli/issues/4302

PR #38, opened 2017-09-05
Improve 'wp plugin' command description

mitraval192:
added command description


mitraval192:
Related to #36

miya0001:
Ah, test is failed, but it looks caused by trunk version of WordPress core.
We need to look into the reason later.
https://travis-ci.org/wp-cli/extension-command/jobs/272039528

PR #39, opened 2017-09-06
Improve command descriptions; add Handbook links

hearvox:
Related: [/wp-cli/issues/4302](https://github.com/wp-cli/wp-cli/issues/4302)

Changes:
___
Improve 'wp plugin' command description in:
https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php 
Current description: Manage plugins.

New description:
Manages site plugins, including installs, activations, and updates.

See the WordPress [Plugin Handbook](https://developer.wordpress.org/plugins/) developer resource.

___
Improve 'wp theme' command description in:
https://github.com/wp-cli/extension-command/blob/master/src/Theme_Command.php
Current description: Manage themes.

New description:
Manages site themes, including installs, activations, and updates.

See the WordPress [Theme Handbook](https://developer.wordpress.org/themes/) developer resource.

PR #41, opened 2017-09-07
change downloading message for theme and plugin in features

miya0001:
Some messages of the WordPress core were changed on trunk, then our tests were broken.
https://travis-ci.org/wp-cli/extension-command/jobs/272525924

Related #38, #39 

See https://core.trac.wordpress.org/ticket/41620

### DIFF
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -5,8 +5,12 @@ Feature: Install WordPress plugins
 
     When I run `wp plugin install https://github.com/runcommand/one-time-login/archive/master.zip --activate`
     Then STDOUT should contain:
+      """"
+      Downloading install
+      """"
+    And STDOUT should contain:
       """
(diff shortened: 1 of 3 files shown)


features/plugin-install.feature
```diff
@@ -6,7 +6,7 @@ Feature: Install WordPress plugins
     When I run `wp plugin install https://github.com/runcommand/one-time-login/archive/master.zip --activate`
     Then STDOUT should contain:
       """
-      Downloading install package from https://github.com/runcommand/one-time-login/archive/master.zip
+      from https://github.com/runcommand/one-time-login/archive/master.zip
```
gitlost: Could we maybe add a `Then STDOUT should contain` step with `Downloading install` message as well here, and keep `package` in front of `from`? To emphasize that we expect a download....

miya0001: Ah, `installation` contains `install` ... right. 😄 
Thanks!

PR #55, opened 2017-10-07
Include dropins like `object-cache.php` when listing installed plugins

### LINKED ISSUE #18: List dropin plugins in wp plugin list
I just had a need to find out whether the `db.php` dropin plugin was in place on a site, but couldn't do it via WP-CLI because dropins aren't listed as part of `wp plugin list`.

I wonder if the command should behave in the same way as the Plugins screen in the admin area, whereby you have to go to the Dropins tab to see dropin plugins. The WP-CLI equivalent would be that you need to pass the dropins status flag to see dropins listed:

    wp plugin list --status=dropins

BhargavBhandari90:
Use 'wp plugin list --status=dropins' for that

Fixes #18

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -511,6 +511,12 @@ Feature: Manage WordPress plugins
     And I run `if test -d wp-content/plugins; then echo "fail"; fi`
     Then STDOUT should be empty
 
+    When I try `wp plugin list --status=dropin`
+    Then STDERR should be:
+      """
+      Error: No plugins found.
+      """
+
     When I run `wp plugin install query-monitor --activate`
     Then STDOUT should not be empty
 
@@ -519,10 +525,7 @@ Feature: Manage WordPress plugins
       | name          | status   |
       | query-monitor | active   |
 
-    When I run `wp plugin list --status=dropin`
+    When I run `wp plugin list --status=dropin --fields=name,title,description`
     Then STDOUT should be a table containing rows:
-      | name                                  | description              |
-      | db.php - Query Monitor Database Class | Custom database class.   |
-
-
-
+      | name   | title                               | description              |
+      | db.php | Query Monitor Database Class        | Custom database class.   |
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -511,12 +511,6 @@ Feature: Manage WordPress plugins
     And I run `if test -d wp-content/plugins; then echo "fail"; fi`
     Then STDOUT should be empty
 
-    When I try `wp plugin list --status=dropin`
-    Then STDERR should be:
-      """
-      Error: No plugins found.
-      """
-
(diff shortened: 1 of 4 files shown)


src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -618,5 +624,43 @@ protected function _search( $args, $assoc_args ) {
 	protected function get_formatter( &$assoc_args ) {
 		return new \WP_CLI\Formatter( $assoc_args, $this->obj_fields, $this->item_type );
 	}
+
+	/**
+	 * Get Dropins.
+	 *
+	 * @param array $assoc_args Data passed in from command.
+	 */
+	protected function get_wp_cli_dropin( $assoc_args ) {
```
danielbachhuber: This method name can be simplified to `get_dropins()`

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -618,5 +624,43 @@ protected function _search( $args, $assoc_args ) {
...
+
+			// Set dropin description.
+			if ( empty( $dropin['Description'] ) ) {
+				$dropin['description'] = $wp_dropins[ $key ][0];
+			}
+
+			$items[] = $dropin;
+		}
+
+		WP_CLI\Utils\format_items( $format, $items, array( 'name', 'description' ) );
```
danielbachhuber: Let's return the `$items` for rendering by the existing code path.

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -618,5 +624,43 @@ protected function _search( $args, $assoc_args ) {
...
+
+			// Get dropin name.
+			$dropin['name'] = $key . ' - ' . $dropin['Name'];
+
+			// Set dropin description.
+			if ( empty( $dropin['Description'] ) ) {
+				$dropin['description'] = $wp_dropins[ $key ][0];
+			}
+
+			$items[] = $dropin;
```
danielbachhuber: Each drop-in should implement all of the fields that a normal plugin array has. If the drop-in doesn't have a corresponding value, you can add an empty string as the value.

features/plugin.feature
```diff
@@ -504,5 +504,25 @@ Feature: Manage WordPress plugins
     query-monitor
     """
 
+  Scenario: Show dropins plugin list
+    Given a WP install
+
+    When I run `rm -rf wp-content/plugins`
+    And I run `if test -d wp-content/plugins; then echo "fail"; fi`
+    Then STDOUT should be empty
+
```
danielbachhuber: Can you include a test for the error state too?

danielbachhuber:
I ended up cleaning up this code so we can get this out the door. Thanks again @BhargavBhandari90

PR #62, opened 2017-11-10
Colorize and add a legend to listed drop-ins

### LINKED ISSUE #60: Dropins are missing in label and color maps
When the list of plugins includes one of the newly added dropins, it produces PHP errors because the function to retrieve the color ends up fetching a reference to an inexistent array index `'dropin'`:

https://github.com/wp-cli/extension-command/blob/v1.1.5/src/WP_CLI/CommandWithUpgrade.php#L502-L509

danielbachhuber:
image

Fixes #60

### DIFF
--- a/features/plugin-status.feature
+++ b/features/plugin-status.feature
@@ -1,0 +1,29 @@
+Feature: List the status of plugins
+
+  Scenario: Status should include drop-ins
+    Given a WP install
+    And a wp-debug.php file:
+      """
+      <?php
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> Have a PR to expose PHP notices/warnings when running behat which I'll post one day rather than having to do ad-hoc debug file requires and tags etc...

💯 ❤️  I think of this literally every time I have to submit one of these pull requests. https://github.com/wp-cli/wp-cli/issues/4504

PR #67, opened 2017-11-21
add word wrapping of description field in a 'list' commands

vbaranovskiy-plesk:
Summary

Hello. 
There is no word wrapping for theme list & plugin list commands for 'description' field 
Steps to reproduce

1 Install Wordpress
2 Install wp-cli
3 Try to get info about plugins or themes via get and list commands

ACTUAL RESULTS:
We have a word wrapping ('description' field) in a get command
[user@ bin]# /opt/plesk/php/5.4/bin/php /wp-cli/php/boot-fs.php --path="<path>" plugin get akismet
but it absent in a list command:
[user@ bin]# /opt/plesk/php/5.4/bin/php /wp-cli/php/boot-fs.php --path="<path>" plugin list

EXPECTED RESULTS:
The 'list' commands should wrap words as like it does a 'get' command
https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L775

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -622,7 +622,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 				'version' => $details['Version'],
 				'update_id' => $file,
 				'title' => $details['Name'],
-				'description' => $details['Description'],
+				'description' => wordwrap( $details['Description'] ),
 			);
 		}
 
--- a/src/Theme_Command.php
+++ b/src/Theme_Command.php
@@ -442,7 +442,7 @@ class Theme_Command extends \WP_CLI\CommandWithUpgrade {
 				'version' => $theme->get('Version'),
 				'update_id' => $theme->get_stylesheet(),
 				'title' => $theme->get('Name'),
-				'description' => $theme->get('Description'),
+				'description' => wordwrap( $theme->get('Description') ),
 				'author' => $theme->get('Author'),
 			);
 


src/Plugin_Command.php
```diff
@@ -622,7 +622,7 @@ protected function get_item_list() {
 				'version' => $details['Version'],
 				'update_id' => $file,
 				'title' => $details['Name'],
-				'description' => $details['Description'],
+				'description' => wordwrap( $details['Description'] ),
```
danielbachhuber: Should this be applied when `--format=json`, or only for `--format=table`?

danielbachhuber:
@vbaranovskiy-plesk Could you share a screenshot of the before and after, just so we have that as a point of reference?

src/Plugin_Command.php
danielbachhuber: Ah, I see that `wp plugin get` does it always...

vbaranovskiy-plesk:
wordwrap
no-wordwrap
wordwrap-list

The first picture shows a response of 'plugin get' command (wordwrap also works for json format)
The second & third picture shows a response of 'plugin list' command with wordwrap and without. 

Seems like commands 'get' and 'list' should returns similar description at least (no wordwrap at all or using a wordwrap in a both places). The word wrapping could be useful sometimes so may be better way is to add it to the list command.

src/Plugin_Command.php
vbaranovskiy-plesk: we are using a command with --format=json option

danielbachhuber:
> Seems like commands 'get' and 'list' should return similar description at least (no wordwrap at all or using a wordwrap in a both places). The word wrapping could be useful sometimes so may be better way is to add it to the list command.

Sure, makes sense to me.

Issue #77, opened 2018-01-15
Search command has an incomplete list of fields


ecotechie:
Looking for something else, I realized that several of the fields from plugins_api() are missing from the [documentation](https://developer.wordpress.org/cli/commands/plugin/search/) and [comments](https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L126) for plugin > search.

Even though they do work on with... wp plugin search akismet --fields=
Ones that work and are not on the docs:
sections, ratings, downloaded, downloadlink, last_updated, added, tags, versions, donate_link, banners, icons, active_installs, contributors

Ones that do not work:
reviews, group

As an aside, what I was trying to do is use last_updated with plugin list. Any pointers on where I may look to make this work?

Not sure how to go about helping out with this. Pull request? Not super git-tastic... Tough happy to learn ;)

schlessera:
To start helping with this, you would create a pull request against the docblock of the `PluginCommand`: https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L126-L142

The online reference is generated automatically out of this file.

Here's an overview of how a pull request is created: https://make.wordpress.org/cli/handbook/pull-requests/

ecotechie:
OK, just created a pull request. Hope it's to your standards. Thanks!

schlessera:
Merged, thanks for the PR!

PR #78, opened 2018-01-16
Add missing fields to plugin search subcommand docblock.

ecotechie:
The [dockblock for plugin search](https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L126-L142) was incomplete. I've tested all of the available [fields from plugins_api()](https://developer.wordpress.org/reference/functions/plugins_api/) and updated the ones that work.


ecotechie:
Any reason they are not in alphabetical order? They seem, somewhat, ordered by likely use...

schlessera:
Closes #77

schlessera:
Thanks for the PR, @ecotechie !

PR #83, opened 2018-01-29
Check args if optional and have all option.

### LINKED ISSUE #79: Documentation doesn't state that `wp plugin|theme install` accepts multiple parameters
It's possible to install multiple plugins or themes at once with `wp plugin|theme install`. The only hint to this is the ellipsis following the positional parameter in the help docs. This should be mentioned explicitly in the docs.

### LINKED ISSUE #82: Running `plugin activate` without a plugin slug should throw an error
Original @schlessera https://github.com/wp-cli/entity-command/issues/136

Internal command validation for `wp plugin activate` does not seem to check whether you actually provided a plugin slug.

It should check it has a valid plugin slug to work on and throw an error otherwise.

via @swissspidy in https://github.com/wp-cli/language-command/issues/27#issuecomment-360478383

gitlost:
Fixes #82 

Checks that plugin and theme commands with optional args `[<plugin>...]` (or `[<theme>...]`) and an `--all` option have something to do, using new helper method `check_optional_args_and_all()` in both `Plugin_Command` and `Theme_Command`.

Also documents in the tests but does not fix some outstanding buggish behaviour re non-pluralized message and using both `--version` and `--all` together.

Also closes #79 as it's related and easy to do.

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -250,7 +250,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	}
 
 	/**
-	 * Activates a plugin.
+	 * Activates one or more plugins.
 	 *
 	 * ## OPTIONS
 	 *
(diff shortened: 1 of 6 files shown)


src/Plugin_Command.php
```diff
@@ -610,6 +606,12 @@ protected function install_from_repo( $slug, $assoc_args ) {
...
 	 */
 	public function update( $args, $assoc_args ) {
+		$all = Utils\get_flag_value( $assoc_args, 'all', false );
+
+		if ( ! ( $args = $this->check_optional_args_and_all( $args, $all ) ) ) {
+			return;
+		}
+
 		if ( isset( $assoc_args['version'] ) ) {
 			foreach ( $this->fetcher->get_many( $args ) as $plugin ) {
```
schlessera: The Fetcher class already goes through all this trouble, even checking whether the provided slugs are actually valid.

Wouldn't it be preferable to run the fetcher earlier, and then check whether it came back with a non-empty result?

gitlost: It's simpler this way. It's just checking that args have or have not been given, and giving an appropriate message. It's not interested in checking args, which is left to the existing code.

Issue #75, opened 2018-01-10
Add --all flag to plugin uninstall and delete


danielbachhuber:
`wp plugin update --all` is helpful shorthand to update all plugins. In the same vein, it would be helpful to add `--all` to `wp plugin uninstall` and `wp plugin delete` to make it simple to uninstall/delete all plugins.

schlessera:
Fixed in #84

PR #84, opened 2018-02-12
Add --all flag for plugin uninstall command

BhargavBhandari90:
for #75

### DIFF
--- a/features/plugin-uninstall.feature
+++ b/features/plugin-uninstall.feature
@@ -36,3 +36,12 @@ Feature: Uninstall a WordPress plugin
       Warning: The 'edit-flow' plugin could not be found.
       """
     And the return code should be 0
+
+  Scenario: Uninstall all installed plugin
+    When I run `wp plugin uninstall --all`
+    Then STDOUT should be:
+      """
+      Uninstalled and deleted 'akismet' plugin.
+      Success: Uninstalled 1 of 1 plugins.
+      """
+    And the return code should be 0
--- a/features/plugin-uninstall.feature
+++ b/features/plugin-uninstall.feature
@@ -27,15 +27,12 @@ Feature: Uninstall a WordPress plugin
 
   Scenario: Attempting to uninstall a plugin that doesn't exist
     When I try `wp plugin uninstall edit-flow`
-    Then STDOUT should be:
-      """
-      Success: Plugin already uninstalled.
-      """
     And STDERR should be:
       """
       Warning: The 'edit-flow' plugin could not be found.
+      Error: No plugins uninstalled.
       """
-    And the return code should be 0
+    And the return code should be 1
 
   Scenario: Uninstall all installed plugins
     When I run `wp plugin uninstall --all`
@@ -50,7 +47,7 @@ Feature: Uninstall a WordPress plugin
     When I run the previous command again
     Then STDOUT should be:
       """
-      Success: Plugin already uninstalled.
+      Success: No plugins installed.
       """
 
   Scenario:  Uninstall all installed plugins when one or more activated
@@ -60,9 +57,11 @@ Feature: Uninstall a WordPress plugin
       Success: Activated 2 of 2 plugins.
       """
     And I run `wp plugin uninstall --all`
-    Then STDERR should contain:
+    And STDERR should be:
       """
       Warning: The 'akismet' plugin is active.
+      Warning: The 'hello.php' plugin is active.
+      Error: No plugins uninstalled.
       """
     And the return code should be 1
     And I run `wp plugin uninstall --deactivate --all`
--- a/features/plugin-uninstall.feature
+++ b/features/plugin-uninstall.feature
@@ -42,6 +42,7 @@ Feature: Uninstall a WordPress plugin
     Then STDOUT should be:
       """
       Uninstalled and deleted 'akismet' plugin.
-      Success: Uninstalled 1 of 1 plugins.
+      Uninstalled and deleted 'hello.php' plugin.
+      Success: Uninstalled 2 of 2 plugins.
       """
     And the return code should be 0
--- a/features/plugin-uninstall.feature
+++ b/features/plugin-uninstall.feature
@@ -47,7 +47,7 @@ Feature: Uninstall a WordPress plugin
     When I run the previous command again
     Then STDOUT should be:
       """
-      Success: No plugins installed.
+      Success: No plugins uninstalled.
       """
 
   Scenario:  Uninstall all installed plugins when one or more activated
--- a/features/plugin-uninstall.feature
+++ b/features/plugin-uninstall.feature
@@ -37,7 +37,7 @@ Feature: Uninstall a WordPress plugin
       """
     And the return code should be 0
 
-  Scenario: Uninstall all installed plugin
+  Scenario: Uninstall all installed plugins
     When I run `wp plugin uninstall --all`
     Then STDOUT should be:
       """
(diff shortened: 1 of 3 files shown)


BhargavBhandari90:
@schlessera @danielbachhuber 

I think the test is not failing because of my changes.

It's showing error from `features/plugin-install.feature` and `features/upgradables.feature` which is not changed.

src/Plugin_Command.php
```diff
@@ -820,6 +820,9 @@ public function get( $args, $assoc_args ) {
 	 * : If set, the plugin files will not be deleted. Only the uninstall procedure
 	 * will be run.
 	 *
+	 * [--all]
+	 * : If set, all deactive plugins will be uninstalled.
```
gitlost: Don't need `deactive`, as the `--deactivate` option covers that.

features/plugin-uninstall.feature
```diff
@@ -36,3 +36,13 @@ Feature: Uninstall a WordPress plugin
       Warning: The 'edit-flow' plugin could not be found.
       """
     And the return code should be 0
+
+  Scenario: Uninstall all installed plugin
+    When I run `wp plugin uninstall --all`
+    Then STDOUT should be:
+      """
+      Uninstalled and deleted 'akismet' plugin.
+      Uninstalled and deleted 'hello.php' plugin.
```
gitlost: This will change to `'hello'` - see below re using `check_optional_args_and_all()`.

features/plugin-uninstall.feature
```diff
@@ -36,3 +36,13 @@ Feature: Uninstall a WordPress plugin
       Warning: The 'edit-flow' plugin could not be found.
       """
     And the return code should be 0
+
+  Scenario: Uninstall all installed plugin
```
gitlost: Typo 'plugin' -> 'plugins'.

features/plugin-uninstall.feature
```diff
@@ -36,3 +36,13 @@ Feature: Uninstall a WordPress plugin
...
+    When I run `wp plugin uninstall --all`
+    Then STDOUT should be:
+      """
+      Uninstalled and deleted 'akismet' plugin.
+      Uninstalled and deleted 'hello.php' plugin.
+      Success: Uninstalled 2 of 2 plugins.
+      """
+    And the return code should be 0
```
gitlost: Also should maybe do `When I run the previous command again` step here with STDOUT `Success: No plugins installed.`.

Also add scenario `Uninstall all installed plugins when one or more activated` with one plugin activated and run without and with `--deactivate` flag.

src/Plugin_Command.php
```diff
@@ -828,7 +831,33 @@ public function get( $args, $assoc_args ) {
 	 */
 	public function uninstall( $args, $assoc_args = array() ) {
 		$successes = $errors = 0;
+
+		$all = Utils\get_flag_value( $assoc_args, 'all', false );
+
 		$plugins = $this->fetcher->get_many( $args );
+
+		// If get all parameter from command, then uninstall all deactive plugins.
+		if ( $all ) {
```
gitlost: So this block goes.

src/Plugin_Command.php
```diff
@@ -853,7 +882,7 @@ public function uninstall( $args, $assoc_args = array() ) {
 			$successes++;
 		}
 		if ( ! $this->chained_command ) {
-			Utils\report_batch_operation_results( 'plugin', 'uninstall', count( $args ), $successes, $errors );
+			Utils\report_batch_operation_results( 'plugin', 'uninstall', count( $plugins ), $successes, $errors );
```
gitlost: And this change isn't needed (if the `if ( count( $plugins ) < count( $args ) ) {` logic above is put in).

src/Plugin_Command.php
```diff
@@ -828,7 +831,33 @@ public function get( $args, $assoc_args ) {
 	 */
 	public function uninstall( $args, $assoc_args = array() ) {
 		$successes = $errors = 0;
+
+		$all = Utils\get_flag_value( $assoc_args, 'all', false );
+
```
gitlost: Could you do
```
		if ( ! ( $args = $this->check_optional_args_and_all( $args, $all ) ) ) {
			return;
		}
```
here. This will mean the `if ( $all ) {` block isn't needed.

src/Plugin_Command.php
```diff
@@ -828,7 +831,33 @@ public function get( $args, $assoc_args ) {
...
 		$plugins = $this->fetcher->get_many( $args );
+
```
gitlost: Hmm, the other `--all` commands do this logic after fetching:
```
		if ( count( $plugins ) < count( $args ) ) {
			$errors = count( $args ) - count( $plugins );
		}
```
which looks like it should be done here too but causes for instance `features/plugin-uninstall.feature:28` to fail with `Error: No plugins uninstalled.` instead of succeeding with `Success: Plugin already uninstalled.`. And also the final step of `features/plugin.feature:3`. So this would be a BC break. But maybe the right thing to do anyway...

features/plugin-uninstall.feature
BhargavBhandari90: It's giving me error in testing if I make it `hello`.

features/plugin-uninstall.feature
BhargavBhandari90: @gitlost I am having some problem with this test. Can you please check.

BhargavBhandari90:
@gitlost See this.
screen shot 2018-02-15 at 5 34 01 pm

src/Plugin_Command.php
```diff
@@ -820,15 +820,52 @@ public function get( $args, $assoc_args ) {
...
+		}
+
 		$successes = $errors = 0;
 		$plugins = $this->fetcher->get_many( $args );
+		if ( count( $plugins ) < count( $args ) ) {
+			$errors = count( $args ) - count( $plugins );
+		}
+
+		// If get all parameter from command, then uninstall all deactive plugins.
+		if ( $all ) {
```
gitlost: This block should be removed as it isn't needed after setting the args with `check_optional_args_and_all()`.

features/plugin-uninstall.feature
```diff
@@ -27,12 +27,45 @@ Feature: Uninstall a WordPress plugin
...
+      Success: No plugins installed.
+      """
+
+  Scenario:  Uninstall all installed plugins when one or more activated
+    And I run `wp plugin activate --all`
+    Then STDOUT should contain:
+      """
+      Success: Activated 2 of 2 plugins.
+      """
+    And I run `wp plugin uninstall --all`
```
gitlost: As @schlessera mentioned need this to be `When I try` as it will fail without the `--deactivate` flag. Then have another successful step with the `--deactivate` flag.

gitlost:
The other failures happening are due to the BC break and are mentioned here https://github.com/wp-cli/extension-command/pull/84#discussion_r167972499. So those tests will have to be adjusted (assuming everyone's ok with the BC break).

schlessera:
@BhargavBhandari90 I've added a few change requests.

For the basic tests setup, it should always read like normal sentences:
```
When ___
Then ___
And ___
[ And ___ ]
[ And ___ ]
...

When ___
Then ___
And ___
[ And ___ ]
[ And ___ ]
...
```

features/plugin-uninstall.feature
```diff
@@ -27,12 +27,45 @@ Feature: Uninstall a WordPress plugin
...
+      Uninstalled and deleted 'akismet' plugin.
+      Uninstalled and deleted 'hello' plugin.
+      Success: Uninstalled 2 of 2 plugins.
+      """
+    And the return code should be 0
+
+    When I run the previous command again
+    Then STDOUT should be:
       """
+      Success: No plugins installed.
```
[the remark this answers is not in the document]
BhargavBhandari90: @schlessera I did little different for this. Let me know if this solution is good or not.

schlessera:
Thanks for the PR, @BhargavBhandari90 !

PR #85, opened 2018-02-12
Allow for modern-wordpress redirect; explicit list not search in upgr…

gitlost:
Adjusts plugin slug not matching project name test to allow for redirect (should have https://github.com/wp-cli-test package for this but doing this as a temporary fix so https://github.com/wp-cli/extension-command/pull/84 can progress).

Changes upgradables test to check for not installed by listing the item explicitly (probably what was intended originally) rather than searching and being dependent on the vagaries of the wordpress.org api.

### DIFF
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -114,9 +114,16 @@ Feature: Install WordPress plugins
     Given a WP install
 
     When I try `wp plugin install user-switching user-switching-not-a-plugin`
-    Then STDERR should be:
+    Then STDERR should contain:
       """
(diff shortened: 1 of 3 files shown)


gitlost:
Okay trunk changed to using the `info/1.2` api https://core.trac.wordpress.org/ticket/43192. This probably means that this data-specific check [CommandWithUpgrade.php#L195](https://github.com/wp-cli/extension-command/blob/v1.1.9/src/WP_CLI/CommandWithUpgrade.php#L195) should adjusted to take account of WP version but for the moment I'll just make the tests agnostic.

PR #100, opened 2018-05-27
Introduce wp theme mod list

montu1996:
Added new command `wp theme mod list`.


montu1996:
@schlessera can you please review this PR?

src/Theme_Mod_Command.php
```diff
@@ -121,6 +121,52 @@ public function get( $args = array(), $assoc_args = array() ) {
...
+	 *     +------------------+---------+
+	 *     | key              | value   |
+	 *     +------------------+---------+
+	 *     | background_color | dd3333  |
+	 *     | link_color       | #dd9933 |
+	 *     | main_text_color  | #8224e3 |
+	 *     +------------------+---------+
+	 *
+	 */
+	public function list( $args = array(), $assoc_args = array() ) {
```
schlessera: `list` is a protected keyword in more recent versions of PHP, so you cannot use it as a method name.

Use `list_` instead, and then add the `@subcommand list` annotation to the docblock, similar to how it is done here: https://github.com/wp-cli/extension-command/blob/master/src/Plugin_Command.php#L1006-L1008

src/Theme_Mod_Command.php
```diff
@@ -121,6 +121,52 @@ public function get( $args = array(), $assoc_args = array() ) {
...
+		if ( ! array_key_exists( 'all', $assoc_args ) ) {
+			$assoc_args['all'] = 1;
+		}
+		// If no format was given ,set to default 'table'.
+		if ( ! array_key_exists( 'format', $assoc_args ) ) {
```
schlessera: Setting the default `--format` should be unnecessary, as the default is already provided in https://github.com/wp-cli/extension-command/pull/100/files#diff-03f3f9875c58703b8a8855ca3e5ca77eR138 and will be enforced by the system.

src/Theme_Mod_Command.php
```diff
@@ -121,6 +121,52 @@ public function get( $args = array(), $assoc_args = array() ) {
 
 	}
 
+	/**
+	 * Gets all theme mods.
```
schlessera: Should be `Gets a list of theme mods.` to match similar commands.

src/Theme_Mod_Command.php
```diff
@@ -121,6 +121,52 @@ public function get( $args = array(), $assoc_args = array() ) {
...
+	 *
+	 * ## OPTIONS
+	 *
+	 * [<mod>...]
```
schlessera: This one should be removed, as the filtering is overridden by the `--all` flag.

features/theme-mod-list.feature
```diff
@@ -0,0 +1,38 @@
+Feature: Manage WordPress theme mods list
```
schlessera: The tests should be written with a list of theme mods that actually contain values, otherwise the tests could just succeed by random chance.

schlessera: You can use something like this to pre-populate the list:

```
Background:
  Given a WP install
  And I run `wp theme mod set key_a value_a`
  And I run `wp theme mod set key_b value_b`
```

montu1996: @schlessera Done. can you please check again? :smiley:

features/theme-mod-list.feature
```diff
@@ -0,0 +1,51 @@
...
+    Then STDOUT should be:
+      """
+      ---
+      - 
+        key: key_a
+        value: value_a
+      - 
+        key: key_b
+        value: value_b
+      """
```
schlessera: Code Style: Always end files with a new line to avoid text processing issues on exotic platforms.

src/Theme_Mod_Command.php
```diff
@@ -121,6 +121,47 @@ public function get( $args = array(), $assoc_args = array() ) {
...
+	 *     +------------------+---------+
+	 *     | background_color | dd3333  |
+	 *     | link_color       | #dd9933 |
+	 *     | main_text_color  | #8224e3 |
+	 *     +------------------+---------+
+	 *
+	 * @subcommand list
+	 */
+	public function list_( $args = array(), $assoc_args = array() ) {
+		if ( ! array_key_exists( 'all', $assoc_args ) ) {
```
schlessera: Now that I think about this some more, I'd prefer to have it just always forcefully set `$assoc_args['all']` to `1` here. If you check for existence of the key first, a user could do something like this, which would produce an error:

```
wp theme mod list --no-all
```

So please just remove the `if ( ! array_key_exists( ... ) { }` that wraps the assignment.

features/theme-mod-list.feature
montu1996: Done.....

montu1996: Done

schlessera:
Thanks for the pull request, @montu1996 !

PR #108, opened 2018-07-19
Search: Add plugin or theme's URL on wordpress.org

### LINKED ISSUE #94: Expose the w.org URL in plugin search results
When using `wp plugin search`, there's no way to quickly access to the plugin's w.org page to get more information about the plugin.

An extra field for the URL would be useful.

GaryJones:
After the plugin_api() or theme_api() call is made, update the results with the generated URL for wordpress.org, made up of a hard-coded domain name, the plural of the item type, and the item's slug.

Fixes #94.

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -152,6 +152,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 *     **icons**: Plugin's Icon Image Link
 	 *     **active_installs**: Plugin's Number of Active Installs
 	 *     **contributors**: Plugin's List of Contributors
+	 *     **url**: Plugin's URL on wordpress.org
 	 *
 	 * [--format=<format>]
 	 * : Render output in a particular format.
--- a/src/WP_CLI/CommandWithUpgrade.php
+++ b/src/WP_CLI/CommandWithUpgrade.php
@@ -613,6 +613,11 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
 
 		$items = $api->$plural;
 
+		// Add `url` for plugin or theme on wordpress.org.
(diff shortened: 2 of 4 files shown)


src/Plugin_Command.php
```diff
@@ -152,6 +152,7 @@ public function status( $args ) {
 	 *     **icons**: Plugin's Icon Image Link
 	 *     **active_installs**: Plugin's Number of Active Installs
 	 *     **contributors**: Plugin's List of Contributors
+	 *     **url**: Plugin's URL on wordpress.org
```
schlessera: CS: There's a trailing space here.

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -613,6 +613,11 @@ protected function _search( $args, $assoc_args ) {
 
 		$items = $api->$plural;
 
+		// Add `url` for plugin or theme on wordpress.org.
+		foreach ( $items as $index => $item_object ) {
+			$item_object->url = 'https://wordpress.org/' . $plural . '/' . $item_object->slug . '/';
```
schlessera: I'd prefer this to use substitution:
```
$item_object->url = "https://wordpress.org/{$plural}/{$item_object->slug}/";
```

src/Plugin_Command.php
GaryJones: Good catch. Fixed in the rebased 26862c0.

schlessera:
This currently fails on WP trunk: https://travis-ci.org/wp-cli/extension-command/jobs/406130240#L665-L704

I assume the `plugins_api()` function has changed.

GaryJones:
> This currently fails on WP trunk

Trying a workaround first.

schlessera:
Thanks for the pull request, @GaryJones !

PR #112, opened 2018-07-24
fix mustache fites paths

wojsmol:
redo of #111


wojsmol:
@schlessera This should go to 1.2.2, 1.2.1 is  already released.

Issue #135, opened 2018-09-28
wp plugin delete <plugin-name> not working if path has a space or & in it


swiftavl:
If the path to a WordPress install has ' ' or '&' in the pathname, then the command 'wp plugin delete <plugin-name>' does not work. Note that list, install, activate, deactivate all work as expected. Delete says it deletes it, but it doesn't. Not sure if other characters and/or commands have the same issue.

example
using a fresh install of WordPress
if ABSPATH defined:  '/Users/userName/development sites/myTestSite/'
the following command will fail to delete the plugin.
> wp plugin delete akismet

I can reproduce it using a ' ' or a '&'. I'm sure there are others.
commands that work:
> wp plugin list
> wp plugin install hello-dolly
> wp plugin activate hello-dolly
> wp plugin deactivate hello-dolly
but this one does not work if ' ' or '&' 
> wp plugin delete hello-dolly

Command Line Log
[wpBugTest.txt](https://github.com/wp-cli/extension-command/files/2429050/wpBugTest.txt)

schlessera:
I have looked into the code to see how this is done.

Strangely, while `wp theme delete` [uses the WordPress function `delete_theme()`](https://github.com/wp-cli/extension-command/blob/v2.0.2/src/Theme_Command.php#L778), `wp plugin delete` [does not use the WordPress function `delete_plugins()` and directly uses the shell `rm` command instead](https://github.com/wp-cli/extension-command/blob/v2.0.2/src/Plugin_Command.php#L1122-L1133).

@danielbachhuber Looking through the file history I fail to find a reason why this diverging behavior exists other than for historical reasons. Do you have additional insight here?

schlessera:
Note: Fixing the above bug is very probably only a matter of properly quoting the folder. However, I'd like to find out first whether we shouldn't rather move to using the Core's built-in `delete_plugins()` function instead.

danielbachhuber:
@schlessera The `delete_plugins()` WordPress function makes an external request for filesystem access: https://developer.wordpress.org/reference/functions/delete_plugins/#source

<img width="982" alt="image" src="https://user-images.githubusercontent.com/36432/46364211-059e8100-c62a-11e8-82cf-63708d1c81cd.png">

It looks like this was originally changed in https://github.com/wp-cli/wp-cli/commit/86416ae9521689dea98f4a20bc7eb2d5176a5168

schlessera:
[...]

PR #138, opened 2018-10-03
wp plugin delete <plugin-name> not working if path has a space or & in it

wojsmol:
fixes https://github.com/wp-cli/extension-command/issues/135


swiftavl:
Thanks everyone for figuring it out. I'll try to be more helpful next time around.

PR #139, opened 2018-10-03
Use "installation" instead of "install"

henrywright:
Related to [#4541](https://github.com/wp-cli/wp-cli/issues/4541) in wp-cli.


henrywright:
Tests fail here because an exact match is expected. I'll amend the test

Issue #90, opened 2018-03-13
`plugin update` does not fail on bad file permissions


emersonveenstra:
Running `wp plugin update --all` does not fail with bad file permissions:

```
emerson@orion-web-1:/var/www/sites/example.com $ wp plugin update --all
Enabling Maintenance mode...
Downloading update from https://downloads.wordpress.org/plugin/wordpress-seo.7.0.3.zip...
Using cached file '/home/emerson/.wp-cli/cache/plugin/wordpress-seo-7.0.3.zip'...
Unpacking the update...
Installing the latest version...
Removing the old version of the plugin...
Warning: The update cannot be installed because we will be unable to copy some files. This is usually due to inconsistent file permissions.
[snipped many files]
Plugin update failed.
Disabling Maintenance mode...
+---------------+-------------+-------------+--------+
| name          | old_version | new_version | status |
+---------------+-------------+-------------+--------+
| wordpress-seo | 6.2         | 7.0.3       | Error  |
+---------------+-------------+-------------+--------+
Success: Plugin already updated.

```

For reference:
```
emerson@orion-web-1:/var/www/sites/example.com $ wp cli version
WP-CLI 1.5.0
emerson@orion-web-1:/var/www/sites/example.com $ php -v
PHP 7.2.1-1+ubuntu16.04.1+deb.sury.org+1

jmichaelward:
It appears that there are two main issues here:
1. WP-CLI outputs a "Success" message when there are plugin failures. This is the case whether updating `--all` or updating a specific plugin, as in https://github.com/wp-cli/wp-cli/issues/4929.
2. WP-CLI shows an actual new_version of a plugin when, in fact, the old version is still installed.

The function that outputs the result of the command is `report_batch_operation_results` on line 910 of ./php/utils.php of the WP-CLI framework. That method accepts a `$failures` parameter that is never set if the plugin updater runs into write permissions issues, so output skips to the else condition of that method, where there are also no successes, and thus, the message "Plugin already updated." is output.

To replicate this bug:
1. Make sure you have default file permissions set in WordPress
2. run `wp plugin install jetpack --version=2.2.8` (or any non-current version)
3. Navigate to the plugins directory via the terminal and run `chmod -w jetpack`.
4. run `wp plugin update jetpack`. You will see output similar to to the above, and in the related ticket.

[...]

PR #143, opened 2018-12-09
🐛 FIX: Failed plugin updates not throwing errors

austinginder:
Resolve issue #90. This checks for a null response from bulk_upgrade and assumes a plugin update did not occur. That may include more errors than intended as plugin update errors are not included in response. Refer to https://developer.wordpress.org/reference/classes/plugin_upgrader/.

### DIFF
--- a/features/plugin-update.feature
+++ b/features/plugin-update.feature
@@ -196,11 +196,14 @@ Feature: Update WordPress plugins
     Given a WP install
     And I run `wp plugin install --force akismet --version=4.0`
     And I run `chmod -w wp-content/plugins/akismet`
-    When I try `wp plugin update akismet`
+    And I try `wp plugin update akismet`
+    And save STDERR as {ERROR}
+    And I run `chmod +w wp-content/plugins/akismet`
 
-    Then STDERR should contain:
+    And I run `echo "{ERROR}"`
+    Then STDOUT should contain:
       """
       Error: 
       """
 
-    And I run `chmod +w wp-content/plugins/akismet`
+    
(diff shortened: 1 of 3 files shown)


src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -401,6 +401,9 @@ protected function update_many( $args, $assoc_args ) {
 						'new_version' => $info['update_version'],
 						'status' => $result[ $info['update_id'] ] !== null ? 'Updated' : 'Error',
 					);
+					if ( $result[ $info['update_id'] ] == null ) {
```
schlessera:
```suggestion
					if ( null === $result[ $info['update_id'] ] ) {
```

features/plugin-update.feature
```diff
@@ -191,3 +191,19 @@ Feature: Update WordPress plugins
       """
       Error: Can't find the requested plugin's version 2.5.4 in the WordPress.org plugin repository (HTTP code 404).
       """
+
+  Scenario: Plugin updates that error should not report a success
```
schlessera: This currently breaks on WP versions below v4.7.0, as Akismet v4.1 requires the presence of `WP_REST_Controller`.

I suggest just skipping the test for WP < 4.7.0.

```suggestion
  @require-wp-4.7
  Scenario: Plugin updates that error should not report a success
```

schlessera:
Thanks for the PR, @austinginder !

austinginder:
@schlessera Sure thing and thanks for helping me work through the Behat test.

PR #145, opened 2019-02-13
Added --exclude argument for plugin deactivate

achyuthajoy:
New argument accepts comma separated values and split to array and unsets the specified plugins from $plugins array.


achyuthajoy:
Needs modification. Closing for now

Issue #127, opened 2018-08-28
Command attempts to activate extension after its installation fails


johnbillion:
Given a network connectivity issue when attempting to install a plugin with the `--activate` flag included, the command attempts to activate the plugin even though its installation failed.

It would be a little nicer if activation wasn't attempted if installation fails.

In addition, the error message is doubled up, presumably because WP outputs it and then WP-CLI outputs it too.

Example:

```
wp plugin install woocommerce --activate
Warning: An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="https://wordpress.org/support/">support forums</a>. (WordPress could not establish a secure connection to WordPress.org. Please contact your server administrator.) in /sites/wp/build/wp-admin/includes/plugin-install.php on line 180
Warning: woocommerce: An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="https://wordpress.org/support/">support forums</a>.
Activating 'woocommerce'...
Warning: The 'woocommerce' plugin could not be found.
Error: No plugins installed.
```

PR #146, opened 2019-02-25
Handle extension activation for WP error or extension not found

abhijitrakas:
closes https://github.com/wp-cli/extension-command/issues/127

### DIFF
--- a/src/WP_CLI/CommandWithUpgrade.php
+++ b/src/WP_CLI/CommandWithUpgrade.php
@@ -210,10 +210,10 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
 				}
 			}
 
-			// Check plugin is available or not.
-			$plugin = $this->fetcher->get_many( array( $slug ) );
+			// Check extension is available or not.
+			$extension = $this->fetcher->get_many( array( $slug ) );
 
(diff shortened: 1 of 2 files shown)


src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -210,7 +210,10 @@ public function install( $args, $assoc_args ) {
 				}
 			}
 
-			if ( $result ) {
+			// Check plugin is available or not.
+			$plugin = $this->fetcher->get_many( array( $slug ) );
```
schlessera: This could just as well be a theme here, so better use the term `extension` instead:
```suggestion
			$extension = $this->fetcher->get_many( array( $slug ) );
```

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -210,7 +210,10 @@ public function install( $args, $assoc_args ) {
...
+
+			if ( ! is_wp_error( $result ) && count( $plugin ) > 0 ) {
```
schlessera:
```suggestion
			if ( ! is_wp_error( $result ) && count( $extension ) > 0 ) {
```

abhijitrakas:
@schlessera PR Updated. Thanks.

schlessera:
Thanks for the PR, @abhijitrakas !

PR #157, opened 2019-03-07
Display warning in plugin list info if plugin version is higher than expected

### LINKED ISSUE #142: Enhancement: Plugin and Theme list should alert if the version is HIGHER than found on .org
This is a rare-bear problem, but sometimes hackers will bump the version of a hacked plugin so that their backdoor won't get triggered for update and thus delete itself. If the `wp plugin list` command alerted people to "This version is higher than expected..." that might help spot those faster.

thrijith:
Fixes #142

```
+------------------+----------+---------------------------------------+---------+
| name             | status   | update                                | version |
+------------------+----------+---------------------------------------+---------+
| akismet          | inactive | This version is higher than expected! | 999     |
| woocommerce      | inactive | available                             | 3.4.5   |
+------------------+----------+---------------------------------------+---------+
```

Theme List
```
+-----------------+----------+---------------------------------------+---------+
| name            | status   | update                                | version |
+-----------------+----------+---------------------------------------+---------+
| custom-theme    | inactive | none                                  | 1.0.26  |
| twentynineteen  | inactive | This version is higher than expected! | 1.5     |
| twentyseventeen | active   | available                             | 2.0     |
| twentysixteen   | inactive | none                                  | 1.9     |
+-----------------+----------+---------------------------------------+---------+
```

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -107,6 +107,7 @@ Feature: Manage WordPress plugins
     And STDOUT should be empty
     And the return code should be 1
 
+  @require-wp-4.0
   Scenario: Install a plugin, activate, then force install an older version of the plugin
     Given a WP install
 
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -563,9 +563,12 @@ Feature: Manage WordPress plugins
        */
 
       add_filter( 'site_transient_update_plugins', function( $value ) {
-          if ( isset( $value->no_update['hello-dolly/hello.php'] ) ) {
-                  $value->no_update['hello-dolly/hello.php']->new_version = '1.5';
-          }
+          	if ( isset( $value->response ) ) {
+              unset ( $value->response['hello-dolly/hello.php'] );
+            }
+            if ( isset( $value->no_update ) && isset( $value->no_update['hello-dolly/hello.php'] ) ) {
+              $value->no_update['hello-dolly/hello.php']->new_version = '1.5';
+            }
           return $value;
        } );
       ?>
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -548,3 +548,30 @@ Feature: Manage WordPress plugins
     Then STDOUT should be a table containing rows:
       | name         | title | description                    |
       | db-error.php |       | Custom database error message. |
+
+  @require-wp-4.0
+  Scenario: Validate installed plugin's version.
+    Given a WP installation
+    And I run `wp plugin install hello-dolly`
+    And a wp-content/mu-plugins/test-plugin-update.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Test Plugin Update
+       * Description: Fakes installed plugin's data to verify plugin version mismatch
+       * Author: WP-CLI tests
+       */
+
+      add_filter( 'site_transient_update_plugins', function( $value ) {
+          if ( isset( $value->no_update['hello-dolly/hello.php'] ) ) {
+                  $value->no_update['hello-dolly/hello.php']->new_version = '1.5';
+          }
+          return $value;
+       } );
+      ?>
+      """
+
+    When I run `wp plugin list`
+    Then STDOUT should be a table containing rows:
+      | name               | status   | update                       | version |
+      | hello-dolly        | inactive | version higher than expected | 1.6     |
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -172,7 +172,7 @@ Feature: Manage WordPress plugins
       """
     And STDERR should contain:
       """
-      Error: Only updated 1 of 3 plugins.
+      Error: Only updated 2 of 3 plugins.
       """
     And the return code should be 1
 
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -563,14 +563,15 @@ Feature: Manage WordPress plugins
        */
 
       add_filter( 'site_transient_update_plugins', function( $value ) {
-          	if ( isset( $value->response ) ) {
-              unset ( $value->response['hello-dolly/hello.php'] );
-            }
-            if ( isset( $value->no_update ) && isset( $value->no_update['hello-dolly/hello.php'] ) ) {
-              $value->no_update['hello-dolly/hello.php']->new_version = '1.5';
-            }
+          if ( ! is_object( $value ) ) {
+              return $value;
+          }
+
+          unset( $value->response['hello-dolly/hello.php'] );
+          $value->no_update['hello-dolly/hello.php']->new_version = '1.5';
+
           return $value;
-       } );
+      } );
       ?>
       """
 
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -628,7 +628,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 
 		foreach ( $this->get_all_plugins() as $file => $details ) {
 			$all_update_info = $this->get_update_info();
-			$update_info     = ( null !== $all_update_info->response[$file] ) ? (array) $all_update_info->response[$file] : null;
+			$update_info     = ( isset( $all_update_info->response[$file] ) && null !== $all_update_info->response[$file] ) ? (array) $all_update_info->response[$file] : null;
 			$name            = Utils\get_plugin_name( $file );
 
 			if ( ! isset( $duplicate_names[ $name ] ) ) {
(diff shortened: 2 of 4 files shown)


thrijith:
PR is in draft as I couldn't find an any data of current theme version like plugin. For plugin that data is available in response's `no_update `.   Is there any other option for theme?

schlessera:
@thrijith The only thing I can think of right now is to go through the themes that state they don't provide an update and do an extra query to retrieve the latest version of these themes through the `theme_information` query:
Image 2019-03-16 at 11 23 48 PM

src/Plugin_Command.php
```diff
@@ -645,6 +647,17 @@ protected function get_item_list() {
 				'title' => $details['Name'],
 				'description' => wordwrap( $details['Description'] ),
 			);
+
+			if ( null === $update_info ) {
+
+				// Get all plugin info that don't have an update.
+				$plugin_update_info = ( isset( $all_update_info->no_update[$file] ) ) ? $all_update_info->no_update[$file] : null;
```
schlessera: Superfluous parentheses:
```suggestion
				$plugin_update_info = isset( $all_update_info->no_update[$file] ) ? $all_update_info->no_update[$file] : null;
```

src/Theme_Command.php
```diff
@@ -436,9 +437,19 @@ protected function get_item_list() {
 				$network_enabled = array();
 		}
 
+		$all_update_info = $this->get_update_info();
+		$checked_themes  = ( isset( $all_update_info->checked ) ) ? $all_update_info->checked : array();
```
schlessera: Superfluous parentheses:
```suggestion
		$checked_themes  = isset( $all_update_info->checked ) ? $all_update_info->checked : array();
```

src/Theme_Command.php
```diff
@@ -436,9 +437,19 @@ protected function get_item_list() {
...
+			foreach ( $checked_themes as $slug => $version ) {
+				$theme_version_info[$slug] = $this->is_theme_version_valid( $slug, $version );
+			}
+		}
+
 		foreach ( wp_get_themes() as $key => $theme ) {
 			$file = $theme->get_stylesheet_directory();
-			$update_info = $this->get_update_info( $theme->get_stylesheet() );
+
+			$update_info = ( isset( $all_update_info->response[$theme->get_stylesheet()]) && null !== $all_update_info->response[$theme->get_stylesheet()] ) ? (array) $all_update_info->response[$theme->get_stylesheet()] : null;
```
schlessera: WPCS - spaces around non-literal array indexes:
```suggestion
			$update_info = ( isset( $all_update_info->response[ $theme->get_stylesheet() ] ) && null !== $all_update_info->response[ $theme->get_stylesheet() ] ) ? (array) $all_update_info->response[ $theme->get_stylesheet() ] : null;
```

src/Theme_Command.php
```diff
@@ -899,4 +915,30 @@ private static function get_template_path( $template ) {
 
 		return $template_path;
 	}
+
+	/**
+	 * Check if current version of the theme is higher than the one available at WP.org
```
schlessera: WPCS - end comments with a period:
```suggestion
	 * Check if current version of the theme is higher than the one available at WP.org.
```

src/Theme_Command.php
```diff
@@ -899,4 +915,30 @@ private static function get_template_path( $template ) {
...
+	 * @param string $slug Theme slug.
+	 * @param string $version Theme current version.
+	 *
+	 * @return bool|string
+	 */
+	protected function is_theme_version_valid( $slug, $version ) {
+		// Get Theme Info.
+		$theme_info = themes_api( 'theme_information', array( 'slug' => $slug ) );
+
+		// Return empty string for themes not on WP.org
```
schlessera: WPCS - end comments with a period:
```suggestion
		// Return empty string for themes not on WP.org.
```

src/Theme_Command.php
```diff
@@ -899,4 +915,30 @@ private static function get_template_path( $template ) {
...
+		if ( is_wp_error( $theme_info ) ) {
+			return '';
+		}
+
+		// Compare theme version info.
+		if ( version_compare( $version, $theme_info->version, '>' ) ) {
```
schlessera: `version_compare()` with an operator already returns a boolean value, so you can just return the negation of it:
```php
return ! version_compare( $version, $theme_info->version, '>' );
```

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -471,17 +471,10 @@ protected function has_update( $slug ) {
 	/**
 	 * Get the available update info
 	 *
-	 * @param string $slug The plugin/theme slug
-	 *
-	 * @return array|null
+	 * @return array
```
schlessera: `get_site_transient()` will either return the value of the transient (which seems to be an object here, not an array, but generally `mixed`) or `false` if the transient was not found.
You should either adapt the docblock to be `mixed` or modify the value before returning it.

src/Plugin_Command.php
```diff
@@ -645,6 +647,17 @@ protected function get_item_list() {
...
+
+				// Compare version and update information in plugin list.
+				if ( null !== $plugin_update_info && version_compare( $details['Version'], $plugin_update_info->new_version, '>' ) ) {
+					$items[ $file ]['update'] = 'This version is higher than expected!';
```
schlessera: I think we should tone this down a bit so that it better fits in with the rest of the output:
```suggestion
					$items[ $file ]['update'] = 'version higher than expected';
```

src/Theme_Command.php
```diff
@@ -453,6 +464,11 @@ protected function get_item_list() {
 				'author' => $theme->get('Author'),
 			);
 
+			// Compare version and update information in theme list.
+			if ( isset( $theme_version_info[ $key ] ) && false === $theme_version_info[ $key ] ) {
+				$items[ $file ]['update'] = 'This version is higher than expected!';
```
schlessera: Toning down the message here as well.
```suggestion
				$items[ $file ]['update'] = 'version higher than expected';
```

src/Plugin_Command.php
```diff
@@ -645,6 +647,17 @@ protected function get_item_list() {
 				'title' => $details['Name'],
 				'description' => wordwrap( $details['Description'] ),
 			);
+
+			if ( null === $update_info ) {
+
+				// Get all plugin info that don't have an update.
```
schlessera:
```suggestion
				// Get info for all plugins that don't have an update.
```

thrijith:
@schlessera I have added a test for verifying the change. Please check and let me know if something needs change.

features/plugin.feature
```diff
@@ -547,3 +548,30 @@ Feature: Manage WordPress plugins
...
+      """
+      <?php
+      /**
+       * Plugin Name: Test Plugin Update
+       * Description: Fakes installed plugin's data to verify plugin version mismatch
+       * Author: WP-CLI tests
+       */
+
+      add_filter( 'site_transient_update_plugins', function( $value ) {
+          if ( isset( $value->no_update['hello-dolly/hello.php'] ) ) {
```
schlessera: I think this could potentially cause sporadic test failures where an update will be noticed after the installation step in line 555.

I think it would be better to do the following instead, without a prior check:
- `unset ( $value->response['hello-dolly/hello.php'] );` (in case an actual update was found)
- `$value->no_update['hello-dolly/hello.php']->new_version = '1.5';` to fake the versions as you already did.

This should always reliable trigger the error condition, if I've read the code correctly.

features/plugin.feature
```diff
@@ -547,3 +548,33 @@ Feature: Manage WordPress plugins
...
+    And a wp-content/mu-plugins/test-plugin-update.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Test Plugin Update
+       * Description: Fakes installed plugin's data to verify plugin version mismatch
+       * Author: WP-CLI tests
+       */
+
+      add_filter( 'site_transient_update_plugins', function( $value ) {
```
schlessera: I'm still not happy with the flow here, there's still a logical gap (you might have an existing `$value->no_update['hello-dolly/hello.php']` and just keep it...).

I suggest rewriting this like so (and hope WP won't return even more nonsense in that filter):
```php
add_filter( 'site_transient_update_plugins', function( $value ) {
    if ( ! is_object( $value ) ) {
        return $value;
    }

    unset( $value->response['hello-dolly/hello.php'] );
    $value->no_update['hello-dolly/hello.php']->new_version = '1.5';

    return $value;
} );
```

schlessera:
Alright, let's merge this version now. We'll probably soon find out if we forgot something else now...;)

Thanks for the PR!

wonderWomantotheResue:
Here's what I'm experiencing with wp cli v2.2.0 release 4/26. upon using wp theme list, I now see an error message next to themes by Elegant: MyResume, Minimal, Glow, and BlueSky. I see this:
versionhigherthanexpected

BlueSky | inactive | version higher than expected | 5.0.12 | network

this error message is a false positive. The Elegant theme Blue sky is on current level today, 5.0.12 and does not need updating. Yet, wp cli thinks it needs updating. It happens with 4 of the Elegant themes we have installed on multi site. Something is out of sync. Plus the legend for the new version of wp theme status does not include U -update Available.
themeStatus_Legend


However, regarding the legend missing the definition for U -update Available seems to be appear as expected today. It looks like it was fixed. I just ran wp theme status today and the U = Update Available now exists.

Issue #166, opened 2019-04-16
The "wp theme search" command has not the "page" option 


amieiro:
In the WP-CLI 2.1.0 I was checking the commands "wp plugin search" and "wp theme search". I have found that both have the "--per-page" option (optional number of results to display), but the "--page" option it is only shown in the plugin command.

So, you can execute

```
wp plugin search contact-form-7 --per-page=20 --page=3 --fields=slug,rating
```
but you can't execute

```
wp theme search woocommerce --per-page=20 --page=3 --fields=slug,rating,preview_url
```

It gives an error:

```
Error: Parameter errors:
unknown --page parameter
```

I suppose this is a bug because it is a no sense that you can indicate the size of the page search but you can't indicate the page you want to get.

schlessera:
Fixing this is just a matter of adding the `[--page]` flag to the docblock of the theme search command and then regenerating the README.md file.

amieiro:
I close the issue because the PR that solve it has been merged https://github.com/wp-cli/extension-command/pull/167

PR #167, opened 2019-04-17
Add the `--page` parameter to the `theme search` command

amieiro:
Fix the https://github.com/wp-cli/extension-command/issues/166 bug: the "wp theme search" command has not the "page" option

### DIFF
--- a/src/Theme_Command.php
+++ b/src/Theme_Command.php
@@ -105,7 +105,13 @@ class Theme_Command extends \WP_CLI\CommandWithUpgrade {
 	 *
 	 * <search>
 	 * : The string to search for.
-	 *
+     *
+     * [--page=<page>]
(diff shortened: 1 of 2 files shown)


src/Theme_Command.php
```diff
@@ -105,7 +105,13 @@ public function status( $args ) {
 	 *
 	 * <search>
 	 * : The string to search for.
-	 *
+     *
```
schlessera: You seem to be using spaces for indentation here, instead of tabs.

WP-CLI uses tabs for indentation and spaces for alignment only.

amieiro: My IDE changed it :/
I will resolve it and send the PR

schlessera:
Thanks for the PR, @amieiro !

PR #170, opened 2019-04-20
Implement CS checking based on the `WP_CLI_CS` ruleset

### LINKED ISSUE #168: Adopt and enforce new `WP_CLI_CS` standard
We have a new PHPCS standard for WP-CLI called `WPCliCS` (props @jrfnl). It is part of the `wp-cli/wp-cli-tests` package starting with version v2.1.0.

To adopt & enforce this new standard, the following actions need to be taken for this repository:

- [x] Create a PR that adds a custom ruleset `phpcs.xml.dist` to the repository

    - Add `phpcs.xml.dist` file
    - Adapt `.distignore` to ignore `phpcs.xml.dist` & `phpunit.xml.dist`
    - Adapt `.gitignore` to ignore `phpunit.xml`, `phpcs.xml` & `.phpcs.xml`
    - Require version `^2.1` of the `wp-cli/wp-cli-tests` as a dev dependency
- [ ] Make any required changes to the code that fail the checks from the above ruleset in separate PRs
- [ ] Merge thre ruleset once all required changes have been processed and merged

A sample PR for a simple repository can be seen here: https://github.com/wp-cli/maintenance-mode-command/pull/3/files

Related https://github.com/wp-cli/wp-cli/issues/5179

thrijith:
Add a PHPCS ruleset using the new `WP_CLI_CS` standard.

Fixes #168

Related https://github.com/wp-cli/wp-cli/issues/5179

### DIFF
--- a/src/WP_CLI/CommandWithUpgrade.php
+++ b/src/WP_CLI/CommandWithUpgrade.php
@@ -18,15 +18,23 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
 
 	public function __construct() {
 		// Do not automatically check translations updates after updating plugins/themes.
-		add_action( 'upgrader_process_complete', function() {
-			remove_action( 'upgrader_process_complete', array( 'Language_Pack_Upgrader', 'async_upgrade' ), 20 );
-		}, 1 );
+		add_action(
+			'upgrader_process_complete',
+			function() {
+				remove_action( 'upgrader_process_complete', array( 'Language_Pack_Upgrader', 'async_upgrade' ), 20 );
+			},
+			1
+		);
 
-		add_filter( 'http_request_timeout', function () {
-			return 1 * MINUTE_IN_SECONDS;
-		}, 999 );
+		add_filter(
+			'http_request_timeout',
+			function () {
+				return 1 * MINUTE_IN_SECONDS;
+			},
+			999
+		);
 
-		$this->fetcher = new \WP_CLI\Fetchers\Plugin;
+		$this->fetcher = new \WP_CLI\Fetchers\Plugin();
 	}
 
 	abstract protected function get_upgrader_class( $force );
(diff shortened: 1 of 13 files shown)


src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -195,9 +204,9 @@ public function install( $args, $assoc_args ) {
 					$errors++;
 				} elseif ( is_wp_error( $result ) ) {
 					$key = $result->get_error_code();
-					if ( in_array( $key, array( 'plugins_api_failed', 'themes_api_failed' ) )
-						&& ! empty( $result->error_data[ $key ] ) && in_array( $result->error_data[ $key ], array( 'N;', 'b:0;' ) ) ) {
-						\WP_CLI::warning( "Couldn't find '$slug' in the WordPress.org {$this->item_type} directory." );
+					if ( in_array( $key, [ 'plugins_api_failed', 'themes_api_failed' ], true )
+						&& ! empty( $result->error_data[ $key ] ) && in_array( $result->error_data[ $key ], [ 'N;', 'b:0;' ], true ) ) {
```
thrijith: I am unsure about this check over here, `in_array( $result->error_data[ $key ], [ 'N;', 'b:0;' ], true )`, I tried to install non existent plugin, but was not able to get this result.

schlessera: Here's the commit in question. There's an associated test that should still pass then...

schlessera: Forgot the actual link: https://github.com/wp-cli/extension-command/commit/cf8631cb6d800ae4dfd32221735a93f5667406c0

thrijith: Ok, only the test was changed in https://github.com/wp-cli/extension-command/pull/85 to adopt a new version of API.

schlessera: Leave it in with the strict check for now, It didn't seem to break something up until now, and I don't think the type can be off in this case.

PR #181, opened 2019-05-31
Fix issue with failing test cases due to version mismatch

thrijith:
See [build](https://travis-ci.org/wp-cli/extension-command/builds/537730179?utm_source=github_status&utm_medium=notification).


thrijith:
One test is still failing on `trunk` also this should not have failed in the first place as the scenario should be running on `WP 4.0`

PR #185, opened 2019-06-20
#129 Added `slug` field for plugins list

vladutilie:
Hi,
This fixes #129. Let me know if I can improve this.

A test output of the `plugin list` command is below:
```
$ vendor/bin/wp plugin list --path='C:\\xampp\\htdocs\\cospol.local'
+------------------------------+----------+-----------+---------+
| name                         | status   | update    | version |
+------------------------------+----------+-----------+---------+
| contact-form-7               | inactive | none      | 5.1.3   |
| health-check                 | active   | none      | 1.3.2   |
| one-click-close-comments     | inactive | none      | 2.6     |
| user-switching               | active   | available | 1.5.0   |
| health-check-disable-plugins | must-use | none      | 1.3     |
+------------------------------+----------+-----------+---------+

$ vendor/bin/wp plugin list --path='C:\\xampp\\htdocs\\cospol.local' --field=slug
contact-form-7/wp-contact-form-7.php
health-check/health-check.php
one-click-close-comments/one-click-close-comments.php
user-switching/user-switching.php
health-check-disable-plugins.php

$ vendor/bin/wp plugin list --path='C:\\xampp\\htdocs\\cospol.local' --fields=slug
+-------------------------------------------------------+
| slug                                                  |
+-------------------------------------------------------+
| contact-form-7/wp-contact-form-7.php                  |
| health-check/health-check.php                         |
| one-click-close-comments/one-click-close-comments.php |
| user-switching/user-switching.php                     |
| health-check-disable-plugins.php                      |
+-------------------------------------------------------+

$ vendor/bin/wp plugin list --path='C:\\xampp\\htdocs\\cospol.local' --fields=name,slug
+------------------------------+-------------------------------------------------------+
| name                         | slug                                                  |
+------------------------------+-------------------------------------------------------+
| contact-form-7               | contact-form-7/wp-contact-form-7.php                  |
| health-check                 | health-check/health-check.php                         |
| one-click-close-comments     | one-click-close-comments/one-click-close-comments.php |
| user-switching               | user-switching/user-switching.php                     |
| health-check-disable-plugins | health-check-disable-plugins.php                      |
+------------------------------+-------------------------------------------------------+

$ vendor/bin/wp plugin list --path='C:\\xampp\\htdocs\\cospol.local' --fields=name,slug,version
+------------------------------+-------------------------------------------------------+---------+
| name                         | slug                                                  | version |
+------------------------------+-------------------------------------------------------+---------+
[...]

PR #186, opened 2019-06-20
Add a new optional "file" field for plugin list

### LINKED ISSUE #129: Allow `plugin list` to output plugins slugs
Hello,

It would be great to allow the command `plugin list` to allow `slug` as a value for the fields parameters.

Command: `wp plugin list --fields=slug`

Expected output:

```
+-------------------------------------------+
| slug                                      |
+-------------------------------------------+
| akismet.php                               |
| hello.php                                 |
| iwp-client/init.php                       |
| plugins-load-order/plugins-load-order.php |
+-------------------------------------------+
```

mustafauysal:
"file" parameter represents the plugin file

fixes #129

Props also to @vladutilie for another fix for the same issue in #185.

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -655,7 +655,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 			}
 
 			$duplicate_names[ $name ][] = $file;
-			$items[ $file ] =             [
+			$items[ $file ]             = [
 				'name'           => $name,
 				'status'         => $this->get_status( $file ),
 				'update'         => (bool) $update_info,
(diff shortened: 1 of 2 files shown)


src/Plugin_Command.php
```diff
@@ -653,7 +655,7 @@ protected function get_item_list() {
 			}
 
 			$duplicate_names[ $name ][] = $file;
-			$items[ $file ]             = [
+			$items[ $file ] = [
```
wojsmol:
```suggestion
	$items[ $file ]             = [
```

wojsmol: revert this change

mustafauysal: done 9693df372b720a7f6fb9686a92ba008ad9a80992

PR #149, opened 2019-03-01
Added optional argument --exclude for plugin deactivate --all

achyuthajoy:
Added optional argument `--exclude` for `plugin deactivate --all` where plugin slugs can be specified as comma-separated. 

Objective: Allow users to exclude specific plugins from deactivation.
E.g. `wp plugin deactivate --all --exclude=jetpack,akismet`

Also, the `--exclude` parameter will work only if `--all` is specified. I've added 2 additional tests to check this feature under `features/plugin-deactivate.feature` 

This is my first contribution to WordPress community. Suggestions and modifications to improve this are appreciated. If this can be achieved with less changes, I'm happy to learn about it as well.

PR #195, opened 2019-09-13
Added optional argument --exclude for plugin deactivate

beaucollins:
Rebase of wp-cli/extension-command#149 onto origin/master on behalf of @achyuthajoy

---

Added optional argument --exclude for plugin deactivate --all where plugin slugs can be specified as comma-separated.

Objective: Allow users to exclude specific plugins from deactivation.
E.g. wp plugin deactivate --all --exclude=jetpack,akismet

Also, the --exclude parameter will work only if --all is specified. I've added 2 additional tests to check this feature under features/plugin-deactivate.feature

This is my first contribution to WordPress community. Suggestions and modifications to improve this are appreciated. If this can be achieved with less changes, I'm happy to learn about it as well.


achyuthajoy:
Added `--exclude` argument to `plugin delete` and `plugin uninstall` with behat tests

src/Plugin_Command.php
```diff
@@ -1195,9 +1225,15 @@ private function get_all_plugins() {
 	 *
 	 * @param array $args Passed-in arguments.
 	 * @param bool $all All flag.
+	 * @param string $exclude Comma separated list of plugin slugs.
```
schlessera: `$verb` param seems to be missing here. Not a problem of this PR per se, but makes sense to add it while we're working on this.
```suggestion
	 * @param string $verb    Optional. Verb to use for the messages. Defaults to 'install'.
	 * @param string $exclude Comma separated list of plugin slugs.
```

src/Plugin_Command.php
```diff
@@ -1195,9 +1225,15 @@ private function get_all_plugins() {
...
 	 * @return array Same as $args if not all, otherwise all slugs.
 	 */
-	private function check_optional_args_and_all( $args, $all, $verb = 'install' ) {
+	private function check_optional_args_and_all( $args, $all, $verb = 'install', $exclude = null ) {
+		// Workaround for optional argument $verb
```
schlessera: This workaround shouldn't be needed, the argument has a default value of `'install'` already.

schlessera:
The `plugin activate` command is missing for the PR to cover all of the commands that have an `--all` flag.

danielbachhuber:
Thanks for your work on this, @beaucollins ! Closing in favor of #321

Issue #194, opened 2019-08-24
Support multiple values on "plugin list --status=X"


Athanasius:
## Feature Request

I wish to find all *not* inactive plugins and themes for a site.  If I use "wp plugin list --status=active" I'll miss any that are "active-network", and vice versa. I'm using "--format=json" so can't simply grep the output lines.

Obviously I could pass the JSON output through something like jq as a workaround, but the ability to do something like "wp plugin list --status=active,active-network" would negate the need for the extra processing.  Alternatively "wp plugin --status=!inactive" or the like would also work.

Yes, some care needs to be taken with the separator or negation operator to ensure they're not legal characters in WP plugins (or themes).

schlessera:
I think it makes sense to accept a comma-separated list here for multiple status values.

Would you be interested in providing a pull request for this, @Athanasius ?

Athanasius:
Given my use case "--status=!inactive" would be the more useful, although I'm sure others would have use cases for multiple positive filters.

Unfortunately the whole reason I was using WP CLI in the first place was so as to avoid needing to develop in PHP.  I'm unqualified to provide a patch for this.

schlessera:
`--status=!inactive` will be more difficult to parse and will not as easily be adapted for other flags/commands. Allowing for a set of multiple elements as a first iteration is much more straight-forward in terms of implementation.

> Unfortunately the whole reason I was using WP CLI in the first place was so as to avoid needing to develop in PHP. I'm unqualified to provide a patch for this.

That's perfectly fine! We'll see whether other people might find inspiration here to provide a patch. Given the usefulness and applicability to other commands, I might even consider writing this one myself if I find the time.

xyulex:
I would like to give it a go. It's the first issue I am trying to do, so be patient... :)

schlessera:
@xyulex Yes, no problem. Just let me know if you need assistance.

xyulex:
I have started having a look at it and found where WPCLI return all the status for a certain plugin but can't find where does it do the filtering for showing the final table at the formatter.

schlessera:
The filtering should be happening here:

https://github.com/wp-cli/extension-command/blob/master/src/WP_CLI/CommandWithUpgrade.php#L488-L493

xyulex:
OK, thanks, I just found it.

xyulex:
@schlessera I have a version of it working on my local machine. I have two questions:

- Would you like to get the file before I do a pull request?
[...]

PR #199, opened 2019-10-23
Support multiple values on "plugin list --status=X" #194

xyulex:
Added support for multiple values in plugin list --status (e.g. wp plugin list --status=active,inactive"

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -309,6 +309,13 @@ Feature: Manage WordPress plugins
       | name       | status   | file                |
       | akismet    | active   | akismet/akismet.php |
 
+    When I run `wp plugin list --status=inactive,active --fields=name,status,file`
+    Then STDOUT should be a table containing rows:
+      | name        | status   | file                  |
+      | akismet     | active   | akismet/akismet.php   |
+      | hello-dolly | inactive | hello-dolly/hello.php |
+
+
   Scenario: Install a plugin when directory doesn't yet exist
     Given a WP install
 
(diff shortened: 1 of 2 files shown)


xyulex:
@schlessera  Travis CI is returning some errors. I launched the behat in my computer and they return the same errors in the original repo. Am I missing something?

xyulex:
@schlessera Can I get some support on it?

schlessera:
@xyulex You're logic seems to be off.

First of all, you're replacing the initial logic, instead of adding to it. So while you're trying to add this status filtering, you're breaking the rest of the fields.

Also, after your `explode()`, you should do something like this for deciding when to remove:
```php
if ( ! in_array( $assoc_args[ 'status' ], $statuses, true ) ) {
	unset( $all_items[ $key ];
}
```

Does that make sense?

xyulex:
@schlessera I already did that change in that commit: https://github.com/wp-cli/extension-command/pull/199/commits/edcf08caea4f5f2a92eaf3090e6e5cc0598a3a18

schlessera:
@xyulex No, this is not what I'm referring to. This is the total change so far in this PR:
Image 2019-11-14 at 3 15 20 PM

As you can see, the entire logic is only executed if it encounters the `status` field. For any other field, where it would previously `unset()` the ones that weren' matching, it does not do anything anymore.

You replaced the logic for handling all fields with the logic for handling the status field only.

schlessera:
Thanks for all the effort, @xyulex!

This was fixed via #234

PR #200, opened 2019-11-08
Add second must-use plugin that is added with wp-cli-tests 2.1.6+

schlessera:
A polyfills mu plugin was added to the tests framework and needs to be accounted for.


schlessera:
Test failure unrelated, merging.

PR #211, opened 2020-02-06
Theme listings and updates do not work properly for extended updates

sawanoboly:
To fix this, I made a change to more accurately compare theme versions, with reference to plugin version comparisons.


sawanoboly:
@schlessera @beaucollins @thrijith 

I don't know who to ask, so sending mention from the previous commit log. Could you please review this PR ?

swissspidy:
Which existing issue does this solve? Could we have a test case for this?

sawanoboly:
@swissspidy thank you for response 😃 

We have released the following plugins.

https://github.com/getshifter/shifter-github

The purpose is to support versioning of themes and plugins using Github Releases. If the operation is from the dashboard, it works fine.

However, wp-cli cannot get the version correctly. This behavior only affects the theme.

```
$ wp  theme list
+-------------------------------------+----------+------------------------------+---------+
| name                                | status   | update                       | version |
+-------------------------------------+----------+------------------------------+---------+
| shifter-github-hosting-theme-sample | active   | version higher than expected | 1.0.0   |
| twentytwenty                        | inactive | none                         | 1.1     |
+-------------------------------------+----------+------------------------------+---------+
```

```
wp  theme update --all --dry-run
Warning: shifter-github-hosting-theme-sample: version higher than expected.
```

The change in this PR is to use the version information obtained by the function if who have added a mechanism to obtain the version from outside like our plugin.

schlessera:
@sawanoboly Please add a test case to this PR that triggers that particular error, so that we can assert the error is reproducible AND the change fixes it.

sawanoboly:
Sorry to late reply, we are trying to add a test case for this change, but we did not include a test for a similar change in the plugin, so we are looking for a test to help. 👀 

https://github.com/getshifter/extension-command/commit/0da028248b1c49398298340990725629f25019b7

@thrijith As a reference, is it possible to show the meaning and test case of this change?

thrijith:
@sawanoboly  Here is the test case that is added to check the verification of warning when plugin version is higher than the one available at  https://wordpress.org/

https://github.com/wp-cli/extension-command/blob/0505108163503febcd15d6c0fca6bb1f499b9be4/features/plugin.feature#L558-L605

To verify if your added fix works as expected, you could install the private theme and check if it has the warning or or not.

Please let me know if that clears your query, let me know in case of any concerns.

mt8:
@schlessera @thrijith
I added a test on updating private themes.  
Could you review it? 
Thanks

miya0001:
Hi @mt8 @sawanoboly 

We need test cases that reproduce the problem as @thrijith said like following. 😊 

```
    When I try `wp theme update {THEME_NAME}`
    Then should be empty
    """
    Theme updated successfully.
    """
    And STDOUT should contain:
    """
    Theme updated successfully.
    """
```

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/d6ba872f57c62785f75ebb6d799dbae7 in case this PR is auto-closed or broken in some way.

Issue #217, opened 2020-03-11
wp theme delete --all throws an error all the time


emersonveenstra:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

`wp theme delete --all` will always throw an error saying it didn't delete all themes

**Describe how other contributors can replicate this bug**

- have a WP site with multiple themes
- run `wp theme delete --all`
- The output will say `Error: Only deleted 1 of 2 themes.` (numbers dependent on number of themes in the folder)

**Describe what you would expect as the correct outcome**

The docs say that `--all` will delete all themes except the active one, so it shouldn't be an error when the active theme is not deleted, since there's a separate flag `--force` that will delete the active theme

**Let us know what environment you are running this on**

```
OS:	Linux 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64
Shell:	/bin/bash
PHP binary:	/usr/bin/php7.3
PHP version:	7.3.15-3+ubuntu18.04.1+deb.sury.org+1
php.ini used:	/etc/php/7.3/cli/php.ini
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/var/www/sites/[redacted]
WP-CLI packages dir:	/home/[redacted]/.wp-cli/packages/
WP-CLI global config:
WP-CLI project config:
WP-CLI version:	2.4.0
```

**Provide a possible solution**

I *think* that https://github.com/wp-cli/extension-command/blob/master/src/Theme_Command.php#L839 just needs to be removed, although i don't have the setup to test all scenarios right now.

schlessera:
> I think that /src/Theme_Command.php@master#L839  just needs to be removed, although i don't have the setup to test all scenarios right now.

I think it would be better to add another check and not even write a warning for active themes in the case where the `--all` flag was passed.

So, something like this:
```php
			if ( $this->is_active_theme( $theme ) && ! $force ) {
				if ( ! $all ) {
					WP_CLI::warning( "Can't delete the currently active theme: $theme_slug" );
					$errors++;
				}
				continue;
			}
```

emersonveenstra:
[...]

PR #218, opened 2020-03-11
wp theme delete --all no longer errors on active theme deletion

emersonveenstra:
Fixes issue #217

I'm not sure if only testing on the "Success: Deleted" string is good enough, but i didn't want to assume there will always be 5 themes.


schlessera:
Thanks for the PR, @emersonveenstra !

PR #222, opened 2020-04-03
Missing `--network` parameter for `plugin is-active`

jespervnielsen:
The code checks for the argument "network" but its not possible to add it, since the cli fails because its not declered as a valid param.

This fix add's it as a argument.

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -946,7 +946,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 * : The plugin to check.
 	 *
 	 * [--network]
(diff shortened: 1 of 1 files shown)


src/Plugin_Command.php
```diff
@@ -945,6 +945,9 @@ public function is_installed( $args, $assoc_args = array() ) {
 	 * <plugin>
 	 * : The plugin to check.
 	 *
+	 * [--network]
+	 * : If set, check if plugin is network-activated
```
schlessera, line 949:
```suggestion
	 * : If set, check if plugin is network-activated.
```

PR #234, opened 2020-06-04
[wp-cli#194] Allow for filtering my a multiple field values for the "list" subcommand. WIP.

### LINKED ISSUE #194: Support multiple values on "plugin list --status=X"

zaantar:
Fixes #194 

This impacts "plugin list" and "theme list" commands.

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -309,6 +309,35 @@ Feature: Manage WordPress plugins
       | name       | status   | file                |
       | akismet    | active   | akismet/akismet.php |
 
+  Scenario: List plugin by multiple statuses
+    Given a WP multisite install
+    And a wp-content/plugins/network-only.php file:
+      """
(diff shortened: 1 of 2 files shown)


zaantar:
@schlessera, can you please take a look, when you have a minute, and let me know if this is going in the right direction?

PR #236, opened 2020-06-04
Deny --dry-run usage when specifying a specific version

vsamoletov:


features/theme.feature
```diff
@@ -386,6 +386,16 @@ Feature: Manage WordPress themes
       twentytwelve,inactive,1.0,{UPDATE_VERSION}
       """
 
+  Scenario: When updating a theme --dry-run cannot be used when specifying a specific version.
+    Given a WP install
+
+    When I try `wp theme update --all --version=whatever --dry-run`
+    Then STDERR should be:
+      """
+      Error: --dry-run cannot be used when specifying a specific version.
```
schlessera, line 395: Just noticed there's kind of a double-use of the word specify here.
Maybe just refer to both of the flags directly?
```suggestion
      Error: --dry-run cannot be used together with --version.
```

src/Theme_Command.php
```diff
@@ -721,6 +721,10 @@ public function update( $args, $assoc_args ) {
 			return;
 		}
 
+		if ( isset( $assoc_args['version'] ) && isset( $assoc_args['dry-run'] ) ) {
+			WP_CLI::error( '--dry-run cannot be used when specifying a specific version.' );
```
schlessera, line 725:
```suggestion
			WP_CLI::error( '--dry-run cannot be used together with --version.' );
```

PR #259, opened 2020-09-28
Add commands to manage auto-updates for plugins and themes

schlessera:
This PR adds the following commands:

`wp plugin auto-updates status [<plugin>...] [--all] [--enabled-only] [--disabled-only]
  [--field=<field>]`
`wp plugin auto-updates enable [<plugin>...] [--all] [--disabled-only]`
`wp plugin auto-updates disable [<plugin>...] [--all] [--enabled-only]`
`wp theme auto-updates status [<theme>...] [--all] [--enabled-only] [--disabled-only]
  [--field=<field>]`
`wp theme auto-updates enable [<theme>...] [--all] [--disabled-only]`
`wp theme auto-updates disable [<theme>...] [--all] [--enabled-only]`

Fixes #259

### DIFF
--- a/src/Theme_Command.php
+++ b/src/Theme_Command.php
@@ -1,5 +1,7 @@
 <?php
 
+use WP_CLI\CommandWithUpgrade;
+use WP_CLI\ParseThemeNameInput;
 use WP_CLI\Utils;
 
 /**
@@ -39,7 +41,9 @@ use WP_CLI\Utils;
  *
  * @package wp-cli
  */
-class Theme_Command extends \WP_CLI\CommandWithUpgrade {
+class Theme_Command extends CommandWithUpgrade {
+
+	use ParseThemeNameInput;
 
 	protected $item_type         = 'theme';
 	protected $upgrade_refresh   = 'wp_update_themes';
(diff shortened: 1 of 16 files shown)


src/Theme_Command.php
```diff
@@ -934,7 +864,7 @@ public function list_( $_, $assoc_args ) {
 	 * @param bool $all All flag.
 	 * @return array Same as $args if not all, otherwise all slugs.
 	 */
-	private function check_optional_args_and_all( $args, $all ) {
+	private function check_optional_args_and_all( $args, $all, $verb = 'install' ) {
```
swissspidy, line 867: Isn't this covered by `ParseThemeNameInput` now?

schlessera: Yes, it is, I forgot to remove this one.
Also, the current code structure is awful and needs to be refactored, but I'll keep that for a later date.

schlessera: Resolved via https://github.com/wp-cli/extension-command/pull/259/commits/b3cf3b6b05419f9645fb8574ae34edf14b67dc48

PR #265, opened 2020-10-20
Avoid notice when $update_info is null

msaggiorato:
I've noticed these notices, and thought it was an easy fix.

```
Notice: Trying to access array offset on value of type null in phar:///usr/local/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 662
Notice: Trying to access array offset on value of type null in phar:///usr/local/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 663
```


schlessera (approved):

schlessera:
Thanks for the PR, @msaggiorato !

PR #290, opened 2021-05-24
Fix incorrect examples

johnjago:
In the examples, somehow `activate` shows up after the command `wp auto-updates enable` and the command thinks `activate` is a plugin or theme on which to enable auto-updates. This seems like a mistake since there is an actual plugin/theme after `activate`.

I could be missing something, but the command didn't seem to work when I tried it as shown in the example.


schlessera (approved):

schlessera:
Thanks for the PR, @johnjago !

PR #302, opened 2021-10-11
Replace one-time-login with a sample plugin in the tests

### LINKED ISSUE #301: Fix: failing tests "Branch names should be removed from Github projects" on PHP 5.6
## Bug Report

First reported in https://github.com/wp-cli/extension-command/pull/300#issuecomment-938755216, The behat tests are failing in WordPress 5.6 because the plugin [one-time-login](https://github.com/danielbachhuber/one-time-login/blob/64398c491c8a16efe9adfc3b1497d3845d464a25/one-time-login.php#L79) is being downloaded in the tests which requires PHP 7.1

- Check any GitHub action "Testing" on PHP 5.6 ([example](https://github.com/wp-cli/extension-command/pull/300/checks?check_run_id=3840427437#step:13:64))

No failing tests.

A simple repo, preferably in wp-cli organisation with just the plugin header can be used in the tests so that this error will not happen in future.

Nikschavan:
Fixes #301

Replaces the plugin [one-time-login](https://github.com/danielbachhuber/one-time-login/) with a [sample WordPress plugin](https://github.com/Nikschavan/test-wordpress-plugin)

### DIFF
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -58,7 +58,7 @@ Feature: Install WordPress plugins
   Scenario: Don't attempt to rename ZIPs uploaded to GitHub's releases page
     Given a WP install
 
-    When I run `wp plugin install https://github.com/wp-cli-test/generic-example-plugin/archive/refs/tags/v0.1.0.zip`
+    When I run `wp plugin install https://github.com/wp-cli-test/generic-example-plugin/releases/download/v0.1.0/generic-example-plugin.0.1.0.zip`
     Then STDOUT should contain:
       """
       Plugin installed successfully.


swissspidy:
@schlessera Should this sample plugin live in the wp-cli org so that we can guarantee it exists? In case https://github.com/Nikschavan/test-wordpress-plugin ever gets deleted :-)

Nikschavan:
I was hoping it would be simpler to fork and get this into wp-cli org? I mentioned in the issue #301 it would be preferable to be in this org.

Maybe anyone with capabilities can create a repo here instead and I can use that links in this PR?

wojsmol:
We have https://github.com/wp-cli-test/generic-example-pluginor AFAIK we can transfer @Nikschavan to `wp-cli-test` organisation.

schlessera:
@Nikschavan This is done, the repo now has two published releases, each with a separate ZIP as attached asset.

PR #305, opened 2021-11-18
Add title/description for mu-plugins.

alexstine:
Currently, only name is populated for mu-plugins. This PR will also fill in the title and add the description.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -9,7 +9,7 @@ on:
 # Cancels all previous workflow runs for the same branch that have not yet completed.
 concurrency:
   # The concurrency group contains the workflow name and the branch name.
-  group: ${{ github.workflow }}-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches:
+      - main
       - master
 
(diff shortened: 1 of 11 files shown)


schlessera:
Thanks for the PR, @alexstine !

The code looks good at first glance, but to merge it, we'd need a Behat test as well to ensure it actually does (and keeps doing) what it is supposed to do. Do you care to add such a test?

alexstine:
@schlessera Can you find another contributor to implement the test? I just don't have time at the moment and I'm sure you are probably in the same boat. Maybe one for next CLI meeting?

Thanks.

wojsmol:
@alexstine Changes reverted in https://github.com/wp-cli/extension-command/pull/305/commits/10b75d3255ab349541d51965564d7bb04f3964bf was correct.

alexstine:
@wojsmol Better now? Just refreshed the fork/branch.

wojsmol:
@alexstine LGTM now. Tests are requiring @schlessera  approval to run.

alexstine:
@wojsmol These check fails don't look related to my changes. Maybe worth a re-run?

swissspidy:
Should we also add support for descriptions?

Tests are all passing now btw

alexstine:
Nothing like all green checks. 👍 

If you think it is a good idea, I can try to get descriptions working or can handle in another PR.

swissspidy:
If you're up for it, that would be great. Can be in this PR or a new one, whatever works best for you. Code-wise it should be mostly copy & paste

alexstine:
@swissspidy Done. Not sure if the tests will pass but hopefully so. Gave this a test on my test server, seems solid enough.

swissspidy (approved):

PR #308, opened 2021-12-17
Fix: prevent parent theme removal on child theme delete --all

ovidiul:
This PR will try and add a fix for the issue mentioned here https://github.com/wp-cli/wp-cli/issues/5570

Basically, it will first loop through the list of fetched themes and check if the actual active theme has a parent theme(is child theme) and if so, it will prevent the parent theme from being deleted unintentionally, this. unless `--force` is supplied.

The case when the user specifically deletes the parent theme folder with `wp theme delete parent-theme` is omitted since that could be intentional.

Happy to add a Behat test if this is considered. Thanks

### DIFF
--- a/src/Theme_Command.php
+++ b/src/Theme_Command.php
@@ -771,10 +771,25 @@ class Theme_Command extends CommandWithUpgrade {
 
 		$successes = 0;
 		$errors    = 0;
+
+		$protected_from_delete = array();
+
 		foreach ( $this->fetcher->get_many( $args ) as $theme ) {
 			$theme_slug = $theme->get_stylesheet();
 
 			if ( $this->is_active_theme( $theme ) && ! $force ) {
+				$protected_from_delete[] = $theme_slug;
+
+				if ( wp_get_theme( $theme_slug )->parent() ) {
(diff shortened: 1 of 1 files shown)


schlessera:
Thanks for the PR, @ovidiul !
Yes, a Behat test will be needed for merging this safely. Are you up for adding one?

ovidiul:
> Thanks for the PR, @ovidiul ! Yes, a Behat test will be needed for merging this safely. Are you up for adding one?

Yes, happy to add a Behat test once I figure that one out. 

I saw some PHPCS issues and a commented line which should not have been disabled, so I've updated the PR with those changes. Thanks!

ovidiul:
@schlessera I've added 2 Behat tests for when the child theme is activated.

Let me know if you see further changes needed. Thanks

Sidsector9 (changes requested):
@ovidiul Thanks for the PR!
I've tested the PR and I've encountered the following scenario:

When a child theme is active, WordPress doesn't allow deletion of the parent theme unless the theme is switched.
But WP-CLI at the moment allows deletion of the parent theme even when the child theme is active. Can you make that change? The rest looks good!

ovidiul:
👋  @Sidsector9 change is now in place.

Thank you.

PR #315, opened 2022-03-07
Added suggested fix for ignoring non-stable releases when updating a plugin using the --minor flag

### LINKED ISSUE #314: The "--minor"-flag causes beta-releases to be installed
## Bug Report

When I run `wp plugin update {slug} --minor` then beta-versions are considered for updating, while if I run `wp plugin update {slug}`or `wp plugin update {slug} --patch` then only stable releases are considered.

Not all plugins in the WordPress.org-repository have beta-versions tho. I'm currently working on a plugin called Servebolt Optimizer that uses beta-releases for testing and development purposes. You can find our releases [here](https://plugins.svn.wordpress.org/servebolt-optimizer/tags/).

1. Set up a fresh WordPress install
2. Run `wp plugin install servebolt-optimizer --version=3.2`
3. Run `wp plugin update servebolt-optimizer --dry-run` and inspect that you are presented with a stable release
4. Run `wp plugin update servebolt-optimizer --patch --dry-run` and inspect that you are presented with a stable release
5. Run `wp plugin update servebolt-optimizer --minor --dry-run` and inspect that you are presented with a beta-release

Note: we're rapidly updating our plugins these days, and if we make a new stable release it will cause step 5 to return a stable version. In this case one might need to use a different plugin repository where the latest version is a non-stable release.

I would expect beta-versions and release candidates to be ignored by WP CLI unless being targeted specifically using the "--version"-argument.

```
OS:	Darwin 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64
Shell:	/bin/zsh
PHP binary:	/Applications/MAMP/bin/php/php8.0.8/bin/php
PHP version:	8.0.8
php.ini used:	/Applications/MAMP/bin/php/php8.0.8/conf/php.ini
MySQL binary:	/usr/local/opt/mysql-client/bin/mysql
MySQL version:	mysql  Ver 8.0.27 for macos11.6 on x86_64 (Homebrew)
SQL modes:
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/user/code/wp-cli-bug-report
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
WP-CLI version:	2.6.0
```

PR with suggested fix can be seen [here](https://github.com/wp-cli/extension-command/pull/315). Note that I was not able to add any tests because the plugin being used in the Behat-tests (WordPress Importer) does not have any beta or release candidates in the WordPress.org repository. I also considered to add a new argument to the command so that users could chose stable/beta/rc in terms of which tags/versions to consider during update.

roberts91:
Fixes #314


danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/20ad98877aff6e71d6ffc44151b61c25 in case this PR is auto-closed or broken in some way.

Issue #323, opened 2022-07-20
`wp theme delete --all` incorrectly deletes parent theme


PowerKiKi:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

I have a child theme and its parent. The child is active and the parent is correctly reported as parent:

```
$ wp theme list
+------------------+--------+--------+---------+
| name             | status | update | version |
+------------------+--------+--------+---------+
| north-martouf    | active | none   |         |
| siteorigin-north | parent | none   | 1.19.11 |
+------------------+--------+--------+---------+
```

When I run `wp theme delete --all`, I expect nothing to happen because the child is active, and the parent while not strictly active is required for the child to work. However the parent theme will incorrectly be deleted:

```
$ wp theme delete --all
Deleted 'siteorigin-north' theme.          <=== WRONG
Success: Deleted 1 of 2 themes.
```

Then the child won't even appear in the list (which may or may not be an unrelated issue):

```
$ wp theme list
+------+--------+--------+---------+
| name | status | update | version |
+------+--------+--------+---------+
+------+--------+--------+---------+
```

Of course this breaks the entire site until the parent is reinstalled. And it prevent us to use `wp theme delete --all` as a single command to clean up all our sites.

**Describe how other contributors can replicate this bug**

See above

**Describe what you would expect as the correct outcome**

I expect `wp theme delete --all` to preserve the active theme and its parent, so the site does not break.

**Let us know what environment you are running this on**

```
OS:	Linux 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:00:31 UTC 2022 x86_64
Shell:	/bin/bash
PHP binary:	/usr/bin/php7.4
PHP version:	7.4.30
php.ini used:	/etc/php/7.4/cli/php.ini
MySQL binary:	/usr/bin/mysql
MySQL version:	mysql  Ver 15.1 Distrib 10.5.16-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path: ***REDACTED***
[...]

PR #324, opened 2022-07-20
Avoid deleting parent of active theme unless using `--force`

danielbachhuber:
Fixes https://github.com/wp-cli/extension-command/issues/323
Fixes https://github.com/wp-cli/extension-command/issues/325

### DIFF
--- a/src/WP_CLI/ParseThemeNameInput.php
+++ b/src/WP_CLI/ParseThemeNameInput.php
@@ -159,6 +159,17 @@ trait ParseThemeNameInput {
 	}
 
 	/**
+	 * Check whether a given theme is the active theme parent.
+	 *
+	 * @param string $theme Theme to check.
+	 *
+	 * @return bool Whether the provided theme is the active theme.
+	 */
+	protected function is_active_parent_theme( $theme ) {
+		return $theme->get_stylesheet_directory() === get_template_directory();
+	}
+
+	/**
 	 * Get the available update info.
 	 *
 	 * @return mixed Available update info.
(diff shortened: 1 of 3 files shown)


src/WP_CLI/ParseThemeNameInput.php
```diff
@@ -158,6 +158,17 @@ protected function is_active_theme( $theme ) {
...
 
+	/**
+	 * Check whether a given theme is the active theme parent.
+	 *
+	 * @param string $theme Theme to check.
+	 *
+	 * @return bool Whether the provided theme is the active theme.
+	 */
+	protected function is_active_parent_theme( $theme ) {
+		return $theme->get_stylesheet_directory() === get_template_directory();
```
schlessera, line 169: I don't think this is correct. This would mean "the parent IS the child theme". That doesn't seem to be what you want to check for here.

schlessera (changes requested):

src/WP_CLI/ParseThemeNameInput.php
danielbachhuber: It is a little confusing but it is correct.

The check: is the `$theme` (moina in the test case) the _parent_ of the currently active theme (as defined by `get_template_directory()`).

schlessera: Ah, I misread the code. I thought it used the `get_stylesheet_directory()` function, and not the `WP_Theme::get_stylesheet_directory()` method... NVM.

schlessera (approved):

Issue #74, opened 2018-01-07
Allow theme slug renaming


ailaG:
Please allow theme slug renaming. For example:
wp install theme <my-repo>.zip --slug=foo

I just encountered that need when I used URLs from GitLab. I installed a parent theme, and it was given a weird name that the host probably derives from its commit ID.
Then I tried to install its child theme, and got an error, because it couldn't find its parent theme.
If I could install the parent theme and set its slug, that wouldn't have happened.

danielbachhuber:
> Please allow theme slug renaming. For example:
> wp install theme .zip --slug=foo

How is this different than using `unzip`?

schlessera:
WP-CLI also supports an action that is similar to `unzip`, namely installing a plugin/theme from a ZIP file, instead of from the plugin repository.

As we already have that functionality, we should make sure that it is useful in all of the common scenarios, ZIP files having the wrong slug is a very common scenario, so I think we should add support for that.

danielbachhuber:
Couple other pieces of relevant information:

* WP-CLI already has magical renaming behavior for GitHub URLs: https://github.com/wp-cli/wp-cli/issues/2994
* `wp theme install <theme|zip|url>...` supports multiple sources. A `--slug=<slug>` argument could only work with one theme specified because multiple themes can't be installed to the same directory.

schlessera:
Yes, when I looked through the code and saw the renaming for the Github URLs, I thought we could build the `--slug` so that:
* `--adapt-slug=<name>` will put the theme into a folder `<name>`
* `--adapt-slug` will extract the ZIP, check the plugin/theme header for the name and use that as slug.

The second variation (without argument) could also be used for multiple archives.

IvanGrginovInf:
Hi, any updates on this? Running into same issue when installing plugins from github in CI/CD as zips (have an action which builds the whole plugin and attaches is as `release.zip` for each release). They get extracted as `release-55y616` (even tho the folder name inside the `release.zip` is just `release`.

That means that on each `wp plugin update https://github.com/org-name/projectname/releases/download/1.2.3/release.zip` it will download it into a different folder.

Having either of the two proposed options above would help solve this.

> Yes, when I looked through the code and saw the renaming for the Github URLs, I thought we could build the slug parameter so that:
> 
> * `--adapt-slug=<name>` will put the theme into a folder `<name>`
> * `--adapt-slug` will extract the ZIP, check the plugin/theme header for the name and use that as slug.
> 
[...]

PR #328, opened 2022-09-14
Support `--adapt-slug` to override the default directory when installing zips

borkweb:
The `install` subcommand for both plugins and themes now support the following:

* `wp [plugin|theme] install whatever.zip --adapt-slug=bork` - When installing a zip (whether remote or local), the directory it gets installed to will be the value in `adapt-slug`; in this case, `bork/`'
* `wp [plugin|theme] install whatever.zip --adapt-slug` - When installing a zip (whether remote or local), the plugin or theme name will be fetched from the bootstrap file or style.css respectively, slug-ify it, and then use that as the directory it gets installed to.

Fixes: #74


danielbachhuber:
@borkweb Looks pretty great so far! I kicked off the test suite to see if there are any side effects.

What do you think should happen for `wp [plugin|theme] install whatever.zip whatever2.zip --adapt-slug=bork`? Should we throw an error?

borkweb:
@danielbachhuber - Oh! I that's a great question. Erroring makes a lot of sense.

danielbachhuber:
@borkweb Sounds good. Can you add a test case for that scenario? It looks like there's a coding standards issue that needs to be fixed too.

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/6d013dbff6c939165c4c03c34b6477a1 in case this PR is auto-closed or broken in some way.

PR #332, opened 2022-10-17
Avoid throwing error when excluding missing plugin

### LINKED ISSUE #330: "wp plugin update --all --exclude=[]" fails if exclude list has plugin name that isn't installed
## Bug Report

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [X] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

I was attempting to do a bulk update of some plugins, excluding some. We have a global list of plugins at our company where we generally skip automatic updates and handle them individually. I've run the same command many times without a problem, but this latest time I got an error back.

From my testing, it appears to be from a plugin name that isn't on the site. If I only have plugins in the list that are installed (whether active or not), it works as expected, but because we have a global list, there are sometimes plugins that aren't installed.

```wp plugin update --all --exclude=plugin-name-that-isnt-installed```

I expected it to run updates on plugins and skip any plugins that were in the list, ignoring anything in the list that wasn't actually on the site.

```
wp cli info 
OS:     Linux 5.4.0-1080-gke #86-Ubuntu SMP Fri Aug 5 20:26:48 UTC 2022 x86_64
Shell:  /bin/zsh
PHP binary:     /usr/bin/php7.4
PHP version:    7.4.30
php.ini used:   /etc/php/7.4/cli/php.ini
MySQL binary:   /usr/bin/mysql
MySQL version:  mysql  Ver 15.1 Distrib 10.3.34-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /home/<client name redacted>
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 2.6.0
```

I ran the command once with only installed plugins and it worked, but adding a plugin that wasn't installed threw an error:
wpcli-update

schlessera:
Fixes #330

### DIFF
--- a/features/plugin-activate.feature
+++ b/features/plugin-activate.feature
@@ -115,3 +115,13 @@ Feature: Activate WordPress plugins
       Success: Activated 1 of 1 plugins.
       """
     And the return code should be 0
+
+  Scenario: Excluding a missing plugin should not throw an error
+    Given a WP install
+    And I run `wp plugin activate --all --exclude=missing-plugin`
+    Then STDERR should be empty
+    And STDOUT should contain:
+      """
+      Success:
+      """
+    And the return code should be 0
--- a/features/plugin-deactivate.feature
+++ b/features/plugin-deactivate.feature
@@ -94,3 +94,13 @@ Feature: Deactivate WordPress plugins
       Error: Please specify one or more plugins, or use --all.
       """
     And STDOUT should be empty
+
+  Scenario: Excluding a missing plugin should not throw an error
+    Given a WP install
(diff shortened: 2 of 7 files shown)


schlessera:
Tests on WP 3.7 are failing for an unrelated reason: https://github.com/wp-cli/extension-command/issues/333

PR #341, opened 2022-11-10
Fix failing PHP 5.6 tests

danielbachhuber:


danielbachhuber:
The problem is that we're not getting a download offer for the WordPress Importer because the minimum supported WP version increased to 5.2...

danielbachhuber:
Closing in favor of https://github.com/wp-cli/extension-command/pull/342

Issue #144, opened 2018-12-12
wp theme update testtheme --version=xxx deletes theme if not in WP gallery


cldevs:
**Summary:**
   "wp theme update" deletes theme folder if theme is not in WordPress.org theme directory  

**Steps to reproduce:**
    Create empty folder testtheme in wp-content/themes folder
    wp theme update testtheme --version=2.0

**Environment:**
    wp-cli 2.0.1
    Amazon Linux 2
    PHP 7.2.11
    WP 5.0

**Severity** - High (Deletes data & breaks sites for themes not in WordPress.org theme directory)

**Expected Results:** 
    No change to theme  
    Output: ERROR: Theme not found in WordPress.org theme directory and cannot be updated with WP CLI"

**Actual Results:** 
    Theme folder and all contents are deleted.  Site broken if active theme.
    Outputs as follows:
        Warning: Couldn't find 'testtheme' in the WordPress.org theme directory.
        Error: No themes installed.

schlessera:
This is indeed a bug in the version-specific update code path: https://github.com/wp-cli/extension-command/blob/v2.0.2/src/Theme_Command.php#L678-L684

WP-CLI first deletes the theme unconditionally and only then tries to install it again.

I can see two different ways of solving this:

1. Move the theme instead of deleting it. Once the install succeeded, delete the previous theme, otherwise clean up and move the old one back in place.
2. Start with a check to make sure that the theme is actually available for installation for deleting the old one.

I'd say 1. is the safer bet, as we can also handle unexpected errors that way.

cldevs:
Option 1 sounds good!  If "BackupAddon" and "RestoreAddon" can be made generic in a helper class then this process can be leveraged for any update as I'd guess there are other conditions where an addon may get deleted unintentionally.  The only issue I see is if there isn't enough disk space to store 2 copies of an addon, but an addon-size check and a free space check would mitigate that risk.

schlessera:
Failing to create a backup and throwing an error is still preferable to "delete-first", so I don't see the size argument as a blocker.

cldevs:
Good point.  Whatever approach resolves this issue before breaks occur is best.  

I do want to flag that the thinking behind this case of "delete-first" may be elsewhere in the cli, but I've only seen it happen in themes.  In the version before 2.0 I saw theme deletion when trying to update any theme that wasn't in the gallery, but in current version I've only seen it when using this --version= flag.

[...]

PR #349, opened 2023-01-27
Avoid deleting theme if the specified version cannot be found

danielbachhuber:
Fixes https://github.com/wp-cli/extension-command/issues/144

The "delete before install" behavior was originally added in https://github.com/wp-cli/wp-cli/pull/1510

Interestingly, it doesn't appear necessary and may have been copy paste from the `wp plugin update` code.

Seems like it should be fine to remove if all of the existing tests pass. I moved a few other relevant tests into my new feature file.

### DIFF
--- a/features/theme-update.feature
+++ b/features/theme-update.feature
@@ -16,3 +16,100 @@ Feature: Update WordPress themes
       Error: No themes installed
       """
     And the wp-content/themes/wpclitesttheme directory should exist
+
+  Scenario: Install a theme, then update to a specific version of that theme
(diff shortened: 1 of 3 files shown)


schlessera:
> The "delete before install" behavior was originally added in https://github.com/wp-cli/wp-cli/pull/1510

It was already added prior to that, with https://github.com/wp-cli/wp-cli/commit/d4fa0c697849bbf46f2f095f968f400e3514cf9c

schlessera (approved):

PR #350, opened 2023-02-06
Include optional 'auto_updates' field in plugin and theme lists

### LINKED ISSUE #344: Include additional column (field) "auto-updates" which can be on or off
- See: wp-cli/ideas#171

## Why this extra field?

- I would like to export the list of extensions with **all** their attributes:
  - **Status: Active/inactive** → ✅ This is already exported.
  - **Auto updates: On / off** → ❌ This info is not there explicitly!
  - There is a column `update` with the possible values `none` and `available`.
    - As I understand it, this expressed whether updates are available in the repository or not.
    - But there is no field (I am aware of) which tracks whether the site administrator set auto-updates on or off for a plugin.

## I'd like the attribute "auto updates on/off" to be restored exactly during automatic re-installation

- If I had certain plugins frozen to a particular version with auto-update OFF, I would like them to get a different treatment during re-installation:
  - Plugin which had auto-update OFF, gets restored with `wp plugin install pluginName --version=storedVersion`
  - Plugin which had auto-update ON, gets restored with `wp plugin install pluginName`
    - which is either the storedVersion, if no newer version is available at the time of re-installation
    - or a newer version, if meanwhile available in the repo.

endriu84:
Fixes #344 

This pull request adds an optional field to the `wp plugin list` and `wp theme list` command. With this change, users can now display the `auto_updates` column when listing plugins or themes by using the `--field` option, like this:

 `wp plugin list --field=name,auto_updates` or
 `wp theme list --field=name,auto_updates` --format=json

I've also updated the test suite with a test case for this scenario.

Note: This is my first contribution to this project. I'd greatly appreciate any feedback or suggestions on how I can improve my code and my contribution process. Thank you for considering this pull request!

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -696,3 +696,16 @@ Feature: Manage WordPress plugins
     Then STDOUT should be a table containing rows:
       | name    | title             | description                                    |
       | test-mu | Test mu-plugin    | Test mu-plugin description                     |
+
+  Scenario: Listing plugins should include name and auto_updates
+    Given a WP install
+    When I run `wp plugin list --fields=name,auto_updates`
+    Then STDOUT should be a table containing rows:
+      | name              | auto_updates         |
+      | hello             | off                  |
+
+    When I run `wp plugin auto-updates enable hello`
+    And I try `wp plugin list --fields=name,auto_updates`
+    Then STDOUT should be a table containing rows:
+      | name              | auto_updates         |
+      | hello             | on                   |
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -698,15 +698,15 @@ Feature: Manage WordPress plugins
       | test-mu | Test mu-plugin    | Test mu-plugin description                     |
 
   @require-wp-5.5
-  Scenario: Listing plugins should include name and auto_updates
+  Scenario: Listing plugins should include name and auto_update
     Given a WP install
-    When I run `wp plugin list --fields=name,auto_updates`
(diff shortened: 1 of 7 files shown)


features/plugin.feature
```diff
@@ -696,3 +696,16 @@ Feature: Manage WordPress plugins
     Then STDOUT should be a table containing rows:
       | name    | title             | description                                    |
       | test-mu | Test mu-plugin    | Test mu-plugin description                     |
+
+  Scenario: Listing plugins should include name and auto_updates
```
danielbachhuber, line 700:
```suggestion
  @require-wp-5.5
  Scenario: Listing plugins should include name and auto_updates
```

features/theme.feature
```diff
@@ -685,3 +685,16 @@ Feature: Manage WordPress themes
       Success:
       """
     And the return code should be 0
+
+  Scenario: Listing themes should include auto_updates
```
danielbachhuber, line 689:
```suggestion
  @require-wp-5.5
  Scenario: Listing themes should include auto_updates
```

danielbachhuber (approved):
Great work on this, @endriu84 ! I appreciate you taking the time to write tests, too 😊

Looks pretty solid all of the way through. I made a few tweaks you might like to look through.

endriu84:
Thank You @danielbachhuber  for your review and tweaks. It feels great to have my small contribution merged into the project.

PR #351, opened 2023-02-08
Add format option to wp plugin|theme auto-updates status

mrsdizzie:
Add --format option to `wp (plugin|theme) auto-updates status` as requested by user here: https://wordpress.slack.com/archives/C02RP4T41/p1675804802225149

I looked at replacing direct use of `Formatter` with https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-utils-format-items/ but it does not seem to support the singular `--field` option

https://github.com/wp-cli/wp-cli/blob/89be15b87bb3c496841b2362975fe0d8ddd8ed21/php/utils.php#L334-L338

### DIFF
--- a/features/plugin-auto-updates-status.feature
+++ b/features/plugin-auto-updates-status.feature
@@ -97,3 +97,21 @@ Feature: Show the status of auto-updates for WordPress plugins
       """
       enabled
       """
+
+  @require-wp-5.5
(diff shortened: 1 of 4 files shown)


danielbachhuber (approved):
Thanks @mrsdizzie ! Made one PHPCS tweak

PR #355, opened 2023-03-02
Ensure a stable version is used when `--minor` or `--patch` specified

danielbachhuber:
Fixes https://github.com/wp-cli/extension-command/issues/314

### DIFF
--- a/src/WP_CLI/CommandWithUpgrade.php
+++ b/src/WP_CLI/CommandWithUpgrade.php
@@ -2,6 +2,7 @@
 
 namespace WP_CLI;
 
+use Composer\Package\Version\VersionParser;
 use Composer\Semver\Comparator;
 use Exception;
 use WP_CLI;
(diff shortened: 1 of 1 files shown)


src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -2,6 +2,7 @@
 
 namespace WP_CLI;
 
+use Composer\Package\Version\VersionParser;
```
swissspidy, line 5: Isn't it `Composer\Semver\VersionParser`? Or is that something else? 🤔

danielbachhuber:
```suggestion
use Composer\Semver\VersionParser;
```

Looks like `Composer\Package\Version\VersionParser` simply subclasses `Composer\Semver\VersionParser`.

We can use the parent class though

PR #364, opened 2023-06-08
Default to false when the "network" flag is omitted in the activate command

### LINKED ISSUE #348: `activate/deactivate` command does not pass `$network_wide` as `bool`
## Bug Report

When calling `wp plugin activate` or `wp plguin deactivate` without explicitly pass `--network`, `$network_wide` will be `null` instead of `false`.

This causes errors if a plugin uses parameter types like this:

```php

function myplugin_activate( bool $network_wide = false ): void {
	// ...
}

register_activation_hook( __FILE__, 'myplugin_activate' );
```

https://github.com/wp-cli/extension-command/blob/266d53d87f9c381677137cc79ed7805c9c0162a5/src/Plugin_Command.php#L311

https://github.com/wp-cli/extension-command/blob/266d53d87f9c381677137cc79ed7805c9c0162a5/src/Plugin_Command.php#LL402C2-L402C2

No errors

```
(Paste the output of "wp cli info" into this box)
```

`$network_wide` should default to `false` so it's always a `bool`

If applicable, add screenshots to help explain (you can just drag&drop images into the Github issue).

raicem:
Fixes #348

For the `activate` command, we're defaulting the `$network_wide` option to  `NULL` if the `--network` flag is omitted. This causes type errors further down in the code, as shown in issue #348.

I thought about doing the same for the `deactivate` command as well, but for deactivation, passing `null` is probably what we should keep since [it has its own meaning](https://github.com/WordPress/WordPress/blob/dbf0e1bf9e4b5632faff61d87ed5d0ed45a8e2bd/wp-admin/includes/plugin.php#L742) in the WordPress' `deactivate_plugins` function.

### DIFF
--- a/features/plugin-activate.feature
+++ b/features/plugin-activate.feature
@@ -62,6 +62,7 @@ Feature: Activate WordPress plugins
       Plugin 'user-switching' activated.
       """
 
+  @require-php-7
   Scenario: Activating a plugin with no network wide option passes down correct types
     Given a wp-content/plugins/example-plugin.php file:
       """
@@ -82,6 +83,7 @@ Feature: Activate WordPress plugins
       Plugin 'example-plugin' activated.
       Success: Activated 1 of 1 plugins.
       """
+      And STDERR should be empty
 
   Scenario: Not giving a slug on activate should throw an error unless --all given
     When I try `wp plugin activate`
--- a/features/plugin-activate.feature
+++ b/features/plugin-activate.feature
@@ -62,6 +62,27 @@ Feature: Activate WordPress plugins
       Plugin 'user-switching' activated.
       """
 
+  Scenario: Activating a plugin with no network wide option passes down correct types
+    Given a wp-content/plugins/example-plugin.php file:
+      """
+      <?php
+      // Plugin Name: Example Plugin
+
+      function example_plugin_activate( bool $network_wide = false ) {
+        // Doesn't matter what we do here, we just need a function definition to check the type
+        return;
+      }
+
+      register_activation_hook( __FILE__, 'example_plugin_activate' );
+      """
+
+    When I run `wp plugin activate example-plugin`
+    Then STDOUT should be:
+      """
+      Plugin 'example-plugin' activated.
+      Success: Activated 1 of 1 plugins.
+      """
+
   Scenario: Not giving a slug on activate should throw an error unless --all given
     When I try `wp plugin activate`
     Then the return code should be 1
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -310,7 +310,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 *     Success: Network activated 1 of 1 plugins.
 	 */
 	public function activate( $args, $assoc_args = array() ) {
-		$network_wide = Utils\get_flag_value( $assoc_args, 'network' );
+		$network_wide = Utils\get_flag_value( $assoc_args, 'network', false );
 		$all          = Utils\get_flag_value( $assoc_args, 'all', false );
 		$all_exclude  = Utils\get_flag_value( $assoc_args, 'exclude' );
 


danielbachhuber (commented):
Can you add feature tests covering this scenario for both `wp plugin activate` and `wp plugin deactivate` ?

raicem:
Yes @danielbachhuber. I struggled to figure out tests. Can I get another week? Otherwise I'll close the PR.

danielbachhuber:
@raicem Yep, take your time. Feel free to stop by the `#cli` channel on [WordPress.org Slack](https://make.wordpress.org/chat/) if you'd like help further.

raicem:
@danielbachhuber Thank you for your patience with this :bow:  I managed to write a test for this regression now. I hope it makes sense! 

While reported in the original issue #348, I can't replicate the type problem with `wp plugin deactivate`. When deactivating, WP CLI doesn't (and probably can't) cause a type error since [the value it passes down to the `deactivate_plugins` function](https://github.com/wp-cli/extension-command/blob/5773bd6c3fce1ff89f596cadbfd81d579c6270a7/src/Plugin_Command.php#L439C48-L439C48) is [not used](https://github.com/WordPress/WordPress/blob/ffa1d288f32e92c3958f2853adf67c2f37be0120/wp-admin/includes/plugin.php#L762) later in the code.

This is not the case when activating a plugin. [The `$network_wide` variable we pass to `activate_plugin`](https://github.com/wp-cli/extension-command/blob/5773bd6c3fce1ff89f596cadbfd81d579c6270a7/src/Plugin_Command.php#L349C3-L349C3) is directly [passed down to the `activate_{$plugin}` hook](https://github.com/WordPress/WordPress/blob/ffa1d288f32e92c3958f2853adf67c2f37be0120/wp-admin/includes/plugin.php#L693) and causes a type error. Therefore I only fixed and added a test for `wp plugin activate`.

features/plugin-activate.feature
```diff
@@ -62,6 +62,27 @@ Feature: Activate WordPress plugins
       Plugin 'user-switching' activated.
       """
 
+  Scenario: Activating a plugin with no network wide option passes down correct types
```
raicem, line 65: Not sure if there is a better `.feature` file to place this test into. It's rather too specific and sticks out from the other feature tests.

swissspidy:
> While reported in the original issue #348, I can't replicate the type problem with `wp plugin deactivate`. When deactivating, WP CLI doesn't (and probably can't) cause a type error since [the value it passes down to the `deactivate_plugins` function](https://github.com/wp-cli/extension-command/blob/5773bd6c3fce1ff89f596cadbfd81d579c6270a7/src/Plugin_Command.php#L439C48-L439C48) is [not used](https://github.com/WordPress/WordPress/blob/ffa1d288f32e92c3958f2853adf67c2f37be0120/wp-admin/includes/plugin.php#L762) later in the code.
> 
> This is not the case when activating a plugin. [The `$network_wide` variable we pass to `activate_plugin`](https://github.com/wp-cli/extension-command/blob/5773bd6c3fce1ff89f596cadbfd81d579c6270a7/src/Plugin_Command.php#L349C3-L349C3) is directly [passed down to the `activate_{$plugin}` hook](https://github.com/WordPress/WordPress/blob/ffa1d288f32e92c3958f2853adf67c2f37be0120/wp-admin/includes/plugin.php#L693) and causes a type error. Therefore I only fixed and added a test for `wp plugin activate`.

Thanks for digging into this!

IIRC I originally ran into this with the `activate` command, but included `deactivate` in the bug report too so that we check whether it's affected as well. So thanks for confirming that the `deactivate` command is not affected!

raicem:
Thank you @swissspidy for the explanation!

My tests failed, probably because the example code in my test doesn't play nicely with PHP 5.6. I'll check that and fix it later today.

features/plugin-activate.feature
swissspidy, line 65:
```suggestion
  @require-php-7
  Scenario: Activating a plugin with no network wide option passes down correct types
```

@raicem This should fix the errors on PHP 5.6

Scalar type hints are only available in PHP 7+

features/plugin-activate.feature
```diff
@@ -62,6 +62,27 @@ Feature: Activate WordPress plugins
...
+      register_activation_hook( __FILE__, 'example_plugin_activate' );
+      """
+
+    When I run `wp plugin activate example-plugin`
+    Then STDOUT should be:
+      """
+      Plugin 'example-plugin' activated.
+      Success: Activated 1 of 1 plugins.
+      """
+
```
swissspidy, line 85:
```suggestion
      And STDERR should be empty

```

Just for completeness, to verify there are no errors

features/plugin-activate.feature
danielbachhuber: I think this should be fine. Thanks for flagging 😊

raicem: Oh, thank you @swissspidy for the suggestion. I have applied this suggestion now.

features/plugin-activate.feature
raicem: Of course. Thanks again! I have applied this change now.

danielbachhuber (approved):
Thanks for your work on this, @raicem !

PR #366, opened 2023-07-25
check if plugin_update_info->new_version is_string to not fail with php version_compare

ui-pschulz:
if update version is not valid you'll get either warning or fatal error, so I'll add is_string to check remote version info to be valid string 

* php versions < 8 you'll see a warning on execution of "wp plugin list"
```
PHP Warning:  version_compare() expects parameter 2 to be string, array given in phar:///.../wp-cli.phar/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 731
```

* php versions >= 8 you'll get fatal error and no further information:
```
thrown in phar:///usr/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php on line 731
Fatal error: Uncaught TypeError: version_compare(): Argument #2 ($version2) must be of type string, array given in phar:///usr/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php:731
Stack trace:
#0 phar:///usr/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php(731): version_compare()
#1 phar:///usr/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php(229): Plugin_Command->get_item_list()
#2 phar:///usr/bin/wp/vendor/wp-cli/extension-command/src/WP_CLI/CommandWithUpgrade.php(502): Plugin_Command->get_all_items()
#3 phar:///usr/bin/wp/vendor/wp-cli/extension-command/src/Plugin_Command.php(1212): WP_CLI\CommandWithUpgrade->_list()
#4 [internal function]: Plugin_Command->list_()
...
```

so I propose this small workaround which should avoid failing with fatal error in case there is no real version information of a plugin

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -727,8 +727,10 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 				// Get info for all plugins that don't have an update.
 				$plugin_update_info = isset( $all_update_info->no_update[ $file ] ) ? $all_update_info->no_update[ $file ] : null;
 
-				// Compare version and update information in plugin list.
-				if ( null !== $plugin_update_info && version_compare( $details['Version'], $plugin_update_info->new_version, '>' ) ) {
+				// Compare version and update information in plugin list if possible
+				if (!is_string($plugin_update_info->new_version)) {
+					$items[ $file ]['update'] = static::INVALID_REMOTE_PLUGIN_VERSION;
+				} else if ( null !== $plugin_update_info && version_compare( $details['Version'], $plugin_update_info->new_version, '>' ) ) {
 					$items[ $file ]['update'] = static::INVALID_VERSION_MESSAGE;
 				}
 			}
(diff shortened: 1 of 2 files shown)


swissspidy:
When can this happen, do you have any steps to reproduce this? Would be great if we could add tests for this.

ui-pschulz:
we have a plugin which interferes with wordpress functionality on getting get_site_transient update_plugins information, to publish other information as well, it looks like the return is not valid, so we ended up on this fatal error. as the error itself does not really help the one executing my suggestion would be to not fail instead point into the direction that something is wrong with retrieving update information. I can not share the plugin easily which produced that issue, it will be fixed in the plugin itself as well, however - might happen to others who hook into get_site_transient plugin_updates mechanism of wordpress.

danielbachhuber:
Filtering the `update_plugins` transient isn't a formal API. I don't think we need to guard against bad data from it. In fact, it's better to have the fatal occur, so you know the bad data is happening and can take action accordingly.

danielbachhuber:
Makes sense, thanks for clarifying @ui-pschulz !

We'll hold off for now. We can always reconsider if it comes up again in the future.

Issue #377, opened 2023-10-21
SQLite Compatibility


swissspidy:
I'm encountering some interesting errors when running tests with SQLite.

For example the plugin activation tests are failing because the SQLite plugin attempts to do a redirect upon activation. That said, I wonder if it should actually already be active or not.

In other cases it's simple the count of plugins that's wrong and needs updating when on SQLite

swissspidy:
* https://github.com/WordPress/sqlite-database-integration/issues/49

swissspidy:
https://github.com/wp-cli/extension-command/actions/runs/6603123825?pr=378 is interesting. Apparently the GitHub Action is running out of disk space with SQLite..

PR #378, opened 2023-10-21
Update some tests for SQLite

swissspidy:
Partially addresses #377. Some scenarios still need updating though.


swissspidy:
Unfortunately there are no logs for the actual "Run Behat" step where the issue with the SQLite tests happens.

danielbachhuber (approved):
Seems kinda suboptimal to duplicate all of these tests 🙃

I don't have a better suggestion at this time, though.

PR #379, opened 2023-11-04
Further improve SQLite testing

swissspidy:
Follow-up to #378

All the remaining failed tests are because of a PHP < 7.3 compatibility bug in the SQLite plugin, see https://github.com/WordPress/sqlite-database-integration/pull/57

One functionality change included (enhancement):

This adds support for filtering themes/plugins by their `auto_update` status.

Related: https://github.com/wp-cli/wp-cli/issues/5859

### DIFF
--- a/features/upgradables.feature
+++ b/features/upgradables.feature
@@ -53,8 +53,8 @@ Feature: Manage WordPress themes and plugins
 
     When I run `wp <type> list`
     Then STDOUT should be a table containing rows:
-      | name   | status   | update    | version   |
-      | <item> | inactive | available | <version> |
+      | name   | status   | update    | version   | auto_update |
+      | <item> | inactive | available | <version> | off         |
 
     When I run `wp <type> list --field=name`
     Then STDOUT should contain:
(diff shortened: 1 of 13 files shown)


features/upgradables.feature
```diff
@@ -203,6 +203,6 @@ Feature: Manage WordPress themes and plugins
     And the return code should be 1
 
     Examples:
-      | type   | type_name | item                    | item_title              | version | zip_file                                                              | file_to_check                                                    |
+      | type   | type_name | item                    | item_title              | version | zip_file                                                               | file_to_check                                                     |
```
schlessera, line 206: What's the deal with this very special space? :)

schlessera (changes requested):

features/upgradables.feature
schlessera: Note: I think this can all be compressed anyway, the free space in-between should not matter.

swissspidy: Huh, I thought the space was off so I added it for better visual alignment locally 🤔

swissspidy: Makes it a bit nicer to look at during local development when everything is nicely aligned:

Screenshot 2023-11-10 at 12 06 16

schlessera: I think that depends on the actual version number, though. But I'm fine with merging it like this.

schlessera (approved):

PR #382, opened 2023-11-10
Support 'wporg_status' and 'wporg_last_updated' as optional `wp plugin list` fields

janw-me:
Fixes [idea#36](https://github.com/wp-cli/ideas/issues/36)

Extends `wp plugin list` with the fields:

- `wp_org` which will get the wp.org status (active, closed or empty)
- `wp_org_updated` which will get the last svn commit date if it's a wp.org plugin.

These fields will take a look at the status of the plugin on wordpress.org and if it ever was a .org plugin it can display the last commit date.

Related https://github.com/wp-cli/wp-cli/issues/5859

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -46,13 +46,13 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	protected $item_type         = 'plugin';
 	protected $upgrade_refresh   = 'wp_update_plugins';
 	protected $upgrade_transient = 'update_plugins';
+	protected $check_wporg       = [ 'status' => false, 'update_date' => false ];
 
 	protected $obj_fields = array(
 		'name',
 		'status',
 		'update',
 		'version',
-		'wordpress.org',
 	);
 
 	/**
@@ -702,6 +702,7 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 			$all_update_info = $this->get_update_info();
 			$update_info     = ( isset( $all_update_info->response[ $file ] ) && null !== $all_update_info->response[ $file ] ) ? (array) $all_update_info->response[ $file ] : null;
 			$name            = Utils\get_plugin_name( $file );
+			$wporg_info      = $this->get_wporg_data( $name );
 
 			if ( ! isset( $duplicate_names[ $name ] ) ) {
 				$duplicate_names[ $name ] = array();
@@ -721,11 +722,11 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 				'file'           => $file,
 				'auto_update'    => in_array( $file, $auto_updates, true ),
 				'author'         => $details['Author'],
-				'wordpress.org'  => $this->get_dotorg_status( $name ),
+				'wp_org'         => $wporg_info['status'],
+				'wp_org_updated' => $wporg_info['last_updated'],
 			];
 
 			if ( null === $update_info ) {
-
 				// Get info for all plugins that don't have an update.
 				$plugin_update_info = isset( $all_update_info->no_update[ $file ] ) ? $all_update_info->no_update[ $file ] : null;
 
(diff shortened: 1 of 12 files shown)


schlessera:
@janw-me the API-specific stuff should be included in the `WpOrgApi` class found in `wp-cli/wp-cli`.

janw-me:
> @janw-me the API-specific stuff should be included in the `WpOrgApi` class found in `wp-cli/wp-cli`.

Fair enough. I'll take a look at it later this weekend.

src/Plugin_Command.php
```diff
@@ -748,6 +758,57 @@ protected function get_item_list() {
...
+	 * Get the wordpress.org status of a plugin.
+	 *
+	 * @param string $plugin_name The plugin slug.
+	 *
+	 * @return string The status of the plugin, includes the last update date.
+	 */
+	protected function get_wporg_data( $plugin_name ) {
+		$data = [
+			'status'       => 'no_wp_org',
+			'last_updated' => '-',
```
danielbachhuber, line 771:
```suggestion
			'last_updated' => '',
```

We can leave this empty when it's not present.

src/Plugin_Command.php
```diff
@@ -748,6 +758,57 @@ protected function get_item_list() {
...
+	/**
+	 * Get the wordpress.org status of a plugin.
+	 *
+	 * @param string $plugin_name The plugin slug.
+	 *
+	 * @return string The status of the plugin, includes the last update date.
+	 */
+	protected function get_wporg_data( $plugin_name ) {
+		$data = [
+			'status'       => 'no_wp_org',
```
danielbachhuber, line 770:
```suggestion
			'status'       => '',
```

We can leave this empty when it's not present.

src/Plugin_Command.php
```diff
@@ -748,6 +758,57 @@ protected function get_item_list() {
...
+		if ( $this->check_wporg ) {
+			$request       = wp_remote_head( 'https://api.wordpress.org/plugins/info/1.0/' . $plugin_name );
+			$response_code = wp_remote_retrieve_response_code( $request );
+			if ( 200 === $response_code ) {
+				$data['status'] = 'active';
+			}
+			// Just because the plugin is not in the api, does not mean it was never on .org.
+		}
+
+		$request       = wp_remote_get( "https://plugins.trac.wordpress.org/log/{$plugin_name}/?limit=1&mode=stop_on_copy&format=rss" );
```
danielbachhuber, line 786: Should we only make this request if we're checking for the date?

danielbachhuber (commented):
Great start!

I'd like to standardize on `wporg_status` and `wporg_last_updated` as the key names.

src/Plugin_Command.php
janw-me: I made a better check in d7ec2741628b9370a9a3229a06a404812c86a654
The answer is, not always,

When the call to api.wp is not succesfull, we need to check the SVN to distinguish  between closed and never on .org.

src/Plugin_Command.php
janw-me: I was not happy about the label itself.  
But personally I prefer when a check like this tells "something". Empty can feel like an unknown failure.

Are there other places where we leave fields empty?

janw-me:
> I'd like to standardize on `wporg_status` and `wporg_last_updated` as the key names.

Relabeling done in: fd9a235c99bc1e6793f90a954e6cbbc424a29f0f

janw-me:
> @janw-me the API-specific stuff should be included in the `WpOrgApi` class found in `wp-cli/wp-cli`.

Fixed in 71a1d231c76f0f76e1387e3e98ed9295c80b22a4
Personally I don't think this is better, as this does the whole request instead of the HEAD.
But I also value consistency , so fine.

src/Plugin_Command.php
```diff
@@ -266,17 +275,19 @@ protected function get_all_items() {
...
+				'status'             => 'dropin',
+				'update'             => false,
+				'update_version'     => null,
+				'update_package'     => null,
+				'update_id'          => '',
+				'file'               => $name,
+				'auto_update'        => false,
+				'author'             => $item_data['Author'],
+				'wporg_status'       => 'no_wp_org',
+				'wporg_last_updated' => '-',
```
swissspidy, line 290: As per another comment, let's leave this empty per default instead 

```suggestion
				'wporg_last_updated' => '',
```

schlessera:
> Personally I don't think this is better, as this does the whole request instead of the HEAD.

@janw-me Note: The `WpOrgApi` class can be adapted to make sure it fulfills all the requirements, of course. It is not considered feature-complete at this point and was meant to be extended/improved as needed.

I'll see how we can adapt it so it does not unnecessarily do the full request.

danielbachhuber:
@janw-me Once your PR is ready, can you also update the PR description with details on how the new feature works? Thanks!

danielbachhuber (approved):
Great work on this, @janw-me !

I made a couple of tweaks (c927a46373c51150188a90d495ec7728efef2574, 9d97c952b411b7561e0bea0ce6a9169c87c5c300) and regenerated the README (1dc730ec48a8b118a447a65c3e1eb814f14f1a74).

Also, as a FYI for the future, please create a feature branch for pull requests as it makes things a bit easier to work with.

janw-me:
> I made a couple of tweaks ([c927a46](https://github.com/wp-cli/extension-command/commit/c927a46373c51150188a90d495ec7728efef2574), [9d97c95](https://github.com/wp-cli/extension-command/commit/9d97c952b411b7561e0bea0ce6a9169c87c5c300)) and regenerated the README ([1dc730e](https://github.com/wp-cli/extension-command/commit/1dc730ec48a8b118a447a65c3e1eb814f14f1a74)).

Great additions.

> Also, as a FYI for the future, please create a feature branch for pull requests as it makes things a bit easier to work with.

Will keep this in mind next time :+1:

Issue #91, opened 2018-03-22
"Latest release" links on GitHub should work with `plugin install`


schlessera:
When using a link to `releases/latest` on Github, Github will redirect to the release page of the latest available release.

The `plugin install` and `theme install` commands should recognize such redirects and properly support extracting the ZIP files out of them. Otherwise, you always have to hardcode a version number into the installation link.

danielbachhuber:
:shipit:

PR #386, opened 2023-11-17
Add support GitHub releases/latest on installation

BogdanUngureanu:
Related to https://github.com/wp-cli/extension-command/issues/91

This PR adds support for installing plugins and themes from GitHub using the `release/latest` URLs.

In order to get the latest release assets, I'm proposing to use the GitHub public API to extract the latest version of the plugin/theme from GH releases. If there's no asset defined, we will fall back to the source code zip file.

~~While the implementation works right now, I haven't added support for GitHub's API request limits. Currently, after 60 requests per hour, the user is required to use an access token in order to make an API request. However, this PR doesn't expose a way to pass that token to the request.~~

**UPDATE**:
I've introduced a new environment variable called _GITHUB_TOKEN_.

### DIFF
--- a/features/plugin-install-github-latest.feature
+++ b/features/plugin-install-github-latest.feature
@@ -1,0 +1,15 @@
+Feature: Install WordPress plugins
+
+  Scenario: Verify that providing a plugin releases/latest GitHub URL will get the latest ZIP
+    Given a WP install
+    When I run `wp plugin install https://github.com/danielbachhuber/one-time-login/releases/latest`
+    Then STDOUT should contain:
+      """
+      Latest release resolved to Version 0.4.0
+      Downloading installation package from
+      """
+    And STDOUT should contain:
+      """
+      Plugin installed successfully.
+      Success: Installed 1 of 1 plugins.
+      """
--- a/src/WP_CLI/ParseGithubUrlInput.php
+++ b/src/WP_CLI/ParseGithubUrlInput.php
@@ -27,8 +27,11 @@ final class ParseGithubUrlInput {
 	 */
 	public function get_the_latest_github_version( $repo_slug ) {
 		$api_url = sprintf( $this->github_releases_api_endpoint, $repo_slug );
+		$token   = getenv( 'GITHUB_TOKEN' );
 
-		$response = \wp_remote_get( $api_url );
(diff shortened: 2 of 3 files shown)


swissspidy:
> * Pass a new optional GitHub access token as a flag on the command.

I'd use an environment variable instead, e.g. `getenv( 'GITHUB_TOKEN' )`

BogdanUngureanu:
Hey @swissspidy, that's a great idea! I've updated the PR to contain the environment variable you proposed. Thanks!

src/WP_CLI/ParseGithubUrlInput.php
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	/**
+	 * Get the GitHub repo from the URL.
+	 *
+	 * @param string $url
+	 *
+	 * @return string|null
+	 */
+	public function get_github_repo_from_url( $url ) {
```
danielbachhuber, line 92:
```suggestion
	public function get_github_repo_from_releases_url( $url ) {
```

Should we include 'releases' in this method name to make it more specific?

src/WP_CLI/ParseGithubUrlInput.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+namespace WP_CLI;
+
+final class ParseGithubUrlInput {
```
danielbachhuber, line 5: Can we move the methods on this class to the existing `CommandWithUpgrade` class? I don't think we want to introduce it as a class others can use, etc.

features/plugin-install-github-latest.feature
```diff
@@ -0,0 +1,15 @@
...
+    Then STDOUT should contain:
+      """
+      Latest release resolved to Version 0.4.0
+      Downloading installation package from
+      """
+    And STDOUT should contain:
+      """
+      Plugin installed successfully.
+      Success: Installed 1 of 1 plugins.
+      """
```
danielbachhuber, line 15: Can you include a test case for the non-happy path too?

danielbachhuber (changes requested):
Looks like a great start, @BogdanUngureanu ! I have a few changes to suggest.

swissspidy:
Closing in favor of #421

PR #393, opened 2024-01-20
Added `--minor` and `--patch` CLI option in `wp theme update`

up1512001:
Fixes https://github.com/wp-cli/extension-command/issues/394

Added missing CLI option for `--minor` and `--patch` into `wp theme update` command list.

## Output Before Adding CLI Options for `--minor` & `--patch`
Screenshot 2024-01-20 at 17 21 30

## Output After Adding CLI Options for `--minor` & `--patch` 
Screenshot 2024-01-20 at 17 21 55

### DIFF
--- a/features/theme-update.feature
+++ b/features/theme-update.feature
@@ -113,3 +113,45 @@ Feature: Update WordPress themes
       """
       Error: Can't find the requested theme's version 1.4.2 in the WordPress.org theme repository (HTTP code 404).
       """
+  
+  Scenario: Error when both --minor and --patch are provided
+    Given a WP install
+
+    When I try `wp theme update --patch --minor --all`
+    Then STDERR should be:
+      """
+      Error: --minor and --patch cannot be used together.
+      """
+    And the return code should be 1
+
+  Scenario: Update a theme to its latest minor release
+    Given a WP install
+    And I run `wp theme install --force twentytwelve --version=2.5.4`
+
+    When I run `wp theme update twentytwelve --minor`
+    Then STDOUT should contain:
+      """
+      Success: Updated 1 of 1 themes.
+      """
+
+    When I run `wp theme get twentytwelve --field=version`
+    Then STDOUT should be:
+      """
+      2.6.1
+      """
+  
+  Scenario: Update a theme to its latest patch release
+    Given a WP install
+    And I run `wp theme install --force twentytwelve --version=0.5`
+
+    When I run `wp theme update twentytwelve --patch`
+    Then STDOUT should contain:
+      """
+      Success: Updated 1 of 1 themes.
+      """
+
+    When I run `wp theme get twentytwelve --field=version`
+    Then STDOUT should be:
+      """
+      0.5.2
+      """
--- a/features/theme-update.feature
+++ b/features/theme-update.feature
@@ -126,7 +126,7 @@ Feature: Update WordPress themes
 
   Scenario: Update a theme to its latest minor release
     Given a WP install
-    And I run `wp theme install --force twentytwelve --version=2.5.4`
+    And I run `wp theme install --force twentytwelve --version=3.0`
 
     When I run `wp theme update twentytwelve --minor`
     Then STDOUT should contain:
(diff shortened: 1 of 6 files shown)


danielbachhuber (changes requested):
Thanks, @up1512001 !

Can you include [some functional tests](https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-and-writing-tests) on this PR?

features/theme-update.feature
```diff
@@ -113,3 +113,45 @@ Feature: Update WordPress themes
...
+
+    When I run `wp theme get twentytwelve --field=version`
+    Then STDOUT should be:
+      """
+      2.6.1
+      """
+  
+  Scenario: Update a theme to its latest patch release
+    Given a WP install
+    And I run `wp theme install --force twentytwelve --version=0.5`
```
swissspidy, line 145: There is no version 0.5 or 0.5.2 of Twenty Twelve 🤔 

See https://themes.trac.wordpress.org/browser/twentytwelve/

There is 1.1 and 1.1.1 though, so these could be used for this test.

features/theme-update.feature
```diff
@@ -113,3 +113,45 @@ Feature: Update WordPress themes
...
+    When I try `wp theme update --patch --minor --all`
+    Then STDERR should be:
+      """
+      Error: --minor and --patch cannot be used together.
+      """
+    And the return code should be 1
+
+  Scenario: Update a theme to its latest minor release
+    Given a WP install
+    And I run `wp theme install --force twentytwelve --version=2.5.4`
```
swissspidy, line 129: There is no version 2.5.4 or 2.6.1 of Twenty Twelve 🤔 

See https://themes.trac.wordpress.org/browser/twentytwelve/

We could use 3.0 with an expected update to 3.9 though

up1512001: @swissspidy I have changed version as you mentioned but still don't know why test cases are failing can you please   help me with this?

swissspidy:
OK so there are a couple of reasons why the tests are not currently working.

First, this PR right now only allows these two new args to be passed, but does not actually change the code to do anything with those args. Support for `theme` item types needs to be added in these two places:

https://github.com/wp-cli/extension-command/blob/955bd947d95ef91da501179c7e561d229ab3cc19/src/WP_CLI/CommandWithUpgrade.php#L351-L357

https://github.com/wp-cli/extension-command/blob/955bd947d95ef91da501179c7e561d229ab3cc19/src/WP_CLI/CommandWithUpgrade.php#L622

I just addressed this in 895e6ad5e406c9ca7b5de3933af5cd53147d1029

Second, and most importantly, the WordPress.org API **by default** does not actually return information about available versions for themes.

Here's the API response for Hello Dolly: https://api.wordpress.org/plugins/info/1.2/?action=plugin_information&request%5Blocale%5D=en_US&request%5Bslug%5D=hello-dolly

Here's the API response for Twenty Twelve: https://api.wordpress.org/themes/info/1.2/?action=theme_information&request%5Blocale%5D=en_US&request%5Bslug%5D=twentytwelve

See how there's no `versions` field.

To address this, we first need to add support for fetching additional fields. After those code modifications, this PR will work as expected.

See https://github.com/wp-cli/wp-cli/pull/5893 for where I add this support.

features/theme-update.feature
swissspidy: See comment below

up1512001:
Hi @swissspidy @danielbachhuber may I know any update regarding [https://github.com/wp-cli/wp-cli/pull/5893](https://github.com/wp-cli/wp-cli/pull/5893) PR and when can I expect feedback for my PR?

swissspidy:
This PR is already in good shape. Once https://github.com/wp-cli/wp-cli/pull/5893 is merged, this PR only requires a dependency update and then it should be good to go.

I just pinged the other committers on https://github.com/wp-cli/wp-cli/pull/5893 to review it.

up1512001:
Hey @swissspidy I just checked the new theme update logic as I understand instead of getting info from style.css we will now fetch from the theme archive for its info am I right?
Thanks for updating it.

swissspidy:
Not sure what you mean, this doesn't really change how themes are looked up or anything.

The one thing I noticed while testing this patch is that `filter_item_list()` stored themes with their full directory path as the key. However, plugins are only stored with their slug. This caused problems when filtering the theme transient, as it turns out that didn't really work before, because we were using the full directory path and not the theme slug. That's why there are these changes to use `get_stylesheet()` instead of `get_stylesheet_directory()` for example.

Another interesting find was that the theme update transient is a list of arrays, whereas for plugins it's a list of objects apparently. So the PR now also caters for that.

As you can see, all existing tests are still passing and the new tests helped uncover these flaws.

swissspidy (approved):

PR #398, opened 2024-02-05
Fix counting of errors when attempting activations

### LINKED ISSUE #396: Incorrect message when there is error in plugin activation
## Bug Report

Lets say plugin `hello.php` as header `Requires PHP: 9.0`

When we run `wp plugin activate hello.php` then following result appears:

```
Warning: Failed to activate plugin. Current PHP version (8.1.16) does not meet minimum requirements for Hello Dolly. The plugin requires PHP 9.0..
Success: Plugin already activated.
```

Message `Success: Plugin already activated.` here is not correct.

Message should be updated to reflect actual situation.

```
OS:     Darwin 22.6.0 Darwin Kernel Version 22.6.0: Tue Nov  7 21:48:06 PST 2023; root:xnu-8796.141.3.702.9~2/RELEASE_X86_64 x86_64
Shell:  /bin/zsh
PHP binary:     /usr/local/Cellar/php@8.1/8.1.27/bin/php
PHP version:    8.1.27
php.ini used:   /usr/local/etc/php/8.1/php.ini
MySQL binary:
MySQL version:
SQL modes:
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /Users/nilambarsharma/Sites/staging/app/public/wp-content/plugins
WP-CLI packages dir:    /Users/nilambarsharma/.wp-cli/packages/
WP-CLI cache dir:       /Users/nilambarsharma/.wp-cli/cache
WP-CLI global config:   /Users/nilambarsharma/.wp-cli/config.yml
WP-CLI project config:  /Users/nilambarsharma/Sites/staging/wp-cli.local.yml
WP-CLI version: 2.9.0
```

schlessera:
Fixes #396

Errors during plugin activation were properly detected, but not actually counted, therefore printing an incorrect `Success: ...` message during the batch operation processing.

By properly counting errors as they are detected, the batch processing will correctly use the success or failure message that is appropriate.

### DIFF
--- a/features/plugin-activate.feature
+++ b/features/plugin-activate.feature
@@ -148,3 +148,26 @@ Feature: Activate WordPress plugins
       Success:
       """
     And the return code should be 0
+
+  Scenario: Incompatible plugins cannot be activated
+    Given a WP installation
+    And a wp-content/plugins/incompatible-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Incompatible Plugin
+       * Requires PHP: 42.0
+       */
+      """
+    And I run `wp cli info | grep "PHP version" | awk '{print $3}'`
+    And save STDOUT as {PHP_VERSION}
+
+    When I try `wp plugin activate incompatible-plugin`
+    Then STDERR should contain:
+      """
+      Failed to activate plugin. Current PHP version ({PHP_VERSION}) does not meet minimum requirements for Incompatible Plugin. The plugin requires PHP 42.0.
+      """
+    And STDOUT should not contain:
+      """
+      Success:
+      """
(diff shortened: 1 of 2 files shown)


features/plugin-activate.feature
```diff
@@ -148,3 +148,27 @@ Feature: Activate WordPress plugins
       Success:
       """
     And the return code should be 0
+
+  @require-wp-5.1
+  Scenario: Incompatible plugins cannot be activated
```
swissspidy, line 153: This is the same as the `Activating a plugin that does not meet PHP minimum throws a warning` on line 108.

Can't we just amend it with the 

```
    And STDOUT should not contain:
      """
      Success:
      """
```

check?

schlessera: Huh! Yes, we can. I did not even check to see if an existing test was present, as I assumed the bug would otherwise not be possible... 😅

schlessera: I reused the existing test now. However, I tweaked the PHP version fetching a bit, as technically the `php` binary and the PHP that WP-CLI uses are not necessarily the same version.

PR #399, opened 2024-02-14
Add —force-check flag to `wp plugin list` and `wp theme list`. (wip)

baizmandesign:
Hi @danielbachhuber,

Here's the work-in-progress pull request that you asked me to submit. Please see [our comment thread on wp-cli/ideas](https://github.com/wp-cli/ideas/issues/183) for the history of our exchange regarding challenges creating tests for this new flag.

Please let me know if I can help in any other way.

Saul

PS. The contents of `features/theme-list.feature` can be ignored and likely needs to be completely rewritten. My plan was to work in `features/plugin-list.feature` and get that right first, then I was going to copy and paste any working code and replace "plugin" with "theme" where needed in `features/theme-list.feature`. But as you know things didn't proceed that far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,18 @@ wp plugin activate [<plugin>...] [--all] [--exclude=<name>] [--network]
     Plugin 'hello' network activated.
     Success: Network activated 1 of 1 plugins.
 
+    # Activate plugins that were recently active.
+    $ wp plugin activate $(wp plugin list --recently-active --field=name)
+    Plugin 'bbpress' activated.
+    Plugin 'buddypress' activated.
+    Success: Activated 2 of 2 plugins.
+
+    # Activate plugins that were recently active on a multisite.
+    $ wp plugin activate $(wp plugin list --recently-active --field=name) --network
+    Plugin 'bbpress' network activated.
+    Plugin 'buddypress' network activated.
+    Success: Activated 2 of 2 plugins.
+
 
 
 ### wp plugin deactivate
@@ -340,7 +352,7 @@ Returns exit code 0 when installed, 1 when uninstalled.
 Gets a list of plugins.
 
 ~~~
-wp plugin list [--<field>=<value>] [--field=<field>] [--fields=<fields>] [--format=<format>] [--status=<status>] [--skip-update-check]
+wp plugin list [--<field>=<value>] [--field=<field>] [--fields=<fields>] [--format=<format>] [--status=<status>] [--skip-update-check] [--recently-active]
 ~~~
 
 Displays a list of the plugins installed on the site with activation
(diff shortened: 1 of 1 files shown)


features/plugin-list.feature
```diff
@@ -0,0 +1,56 @@
+Feature: List WordPress plugins
+
+  Scenario: Refresh update_plugins transient when listing plugins with --force-check flag
```
danielbachhuber, line 3: @baizmandesign What do you think of this Behat test refactor?

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -500,6 +500,11 @@ static function ( $result ) {
 	// phpcs:ignore PSR2.Methods.MethodDeclaration.Underscore -- Whitelisting to provide backward compatibility to classes possibly extending this class.
 	protected function _list( $_, $assoc_args ) {
 
+		// If `--force-check` flag is present, delete the upgrade transient.
+		if ( true === (bool) Utils\get_flag_value( $assoc_args, 'force-check', false ) ) {
+			delete_site_transient( $this->upgrade_transient );
+		}
+
```
danielbachhuber, line 507: What should happen if someone uses `--force-check` and `--skip-update-check` at the same time?

baizmandesign: Hi @danielbachhuber,

That's a good question. As currently written, `--force-check` is toothless if `--skip-update-check` is also passed to the command.

Here are a couple of proposals for resolving this matter (assuming we're solely talking about plugins, as an example):

1. Let the code for both functions execute—delete the plugin transient and skip the call to `wp_update_plugins()`—and issue a warning: "Plugin updates were requested to be force-checked and were not checked due to the --skip-update-check argument."
2. Make it so that the two options are mutually exclusive: you can run either one or the other. Running both results in an error: "Plugins cannot be both force-checked and skipped. Choose one."
3. Let one command always "over-rule" the other. For example, we might decide that `--skip-update-check` should always prevail if both are present. It can proceed quietly (and not delete the plugin transient) or issue a warning ("Plugin updates were not force-checked due to the --skip-update-check argument."). On the other hand, we could decide that `--force-check` always over-rules `--skip-update-check`.

I think I'm partial to option 2 but can be convinced of choosing one of the other options (or alternatives unmentioned above).

Thoughts?

features/plugin-list.feature
baizmandesign: I ran the behat tests, and they worked for me! Very exciting. A couple comments:

+ I'm unclear why the `wp plugin list --fields=name,status,update` command needs to be executed twice after `update-transient.php` is run.
+ Is it an issue that the version of the Hello Dolly plugin, 1.7.2, is hard-coded in the test?

src/WP_CLI/CommandWithUpgrade.php
danielbachhuber: @baizmandesign Option 2 works!

baizmandesign: Sounds good, @danielbachhuber! I'll get to work on implementing the error and the theme version of the flag.

Issue #407, opened 2024-02-17
Plugin Dependencies Support


Luc45:
## Feature Request

WordPress 6.5 adds the [Plugin Dependencies](https://make.wordpress.org/core/2024/02/15/merge-announcement-plugin-dependencies/) capability. It would be useful if `plugin install` had a `--with-dependencies` flag to also install the dependencies of the plugin being installed.

swissspidy:
Generally speaking, we need to test with the new plugin dependencies feature and see if and what to change. If the plugin‘s API provides this information easily, then I could see such a flag being useful for sure.

Luc45:
I haven't looked at how WP Core does it, but it should probably be surfaced on the API already to render it on the wp-admin?plugins.php page

swissspidy:
Right. This is surfaced in the API via the `requires_plugins` field which contains an array of slugs.

Example: https://api.wordpress.org/plugins/info/1.2/?action=plugin_information&request%5Blocale%5D=en_US&request%5Bslug%5D=wpsso-wc-metadata&fields=versions

So should be available for us too 👍

swissspidy:
Listing dependencies in `wp plugin get <slug>` would also be useful

PR #414, opened 2024-04-11
Show additional plugin headers values in `wp plugin get <plugin>` output

ernilambar:
* Plugin header `RequiresPlugins` is also displayed if exists.
* Added feature test for plugin get. One scenario for `RequiresPlugins` and one with basic test in separate file `plugin-get.feature`

Related: https://github.com/wp-cli/extension-command/issues/407

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -947,27 +947,46 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 *     {"name":"bbpress","title":"bbPress","author":"The bbPress Contributors","version":"2.6-alpha","description":"bbPress is forum software with a twist from the creators of WordPress.","status":"active"}
 	 */
 	public function get( $args, $assoc_args ) {
+		$default_fields = array(
+			'name',
+			'title',
+			'author',
+			'version',
+			'description',
+			'status',
+		);
+
 		$plugin = $this->fetcher->get_check( $args[0] );
 		$file   = $plugin->file;
 
 		$plugin_data = get_plugin_data( WP_PLUGIN_DIR . '/' . $file, false, false );
 
 		$plugin_obj = (object) [
-			'name'        => Utils\get_plugin_name( $file ),
-			'title'       => $plugin_data['Name'],
-			'author'      => $plugin_data['Author'],
-			'version'     => $plugin_data['Version'],
-			'description' => wordwrap( $plugin_data['Description'] ),
-			'status'      => $this->get_status( $file ),
+			'name'             => Utils\get_plugin_name( $file ),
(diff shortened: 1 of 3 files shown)


src/Plugin_Command.php
```diff
@@ -961,6 +961,10 @@ public function get( $args, $assoc_args ) {
 			'status'      => $this->get_status( $file ),
 		];
 
+		if ( isset( $plugin_data['RequiresPlugins'] ) && ! empty( $plugin_data['RequiresPlugins'] ) ) {
```
swissspidy, line 964: Should we also add the field if it's empty?

swissspidy:
Makes we wonder whether we shouldn't add more information from `get_plugin_data` to this list by default. For example the `RequiresWP` and `RequiresPHP` information could be useful (both added in 5.3), or `UpdateURI` (added in 5.8)

ernilambar:
@swissspidy We could keep `name`, `title`, `author`, `version`, `description` and `status` as default fields and other as optional `AVAILABLE FIELDS`. And for available optional fields we could add those new fields, like `requires_plugins`, `requires_php`, etc.

src/Plugin_Command.php
```diff
@@ -941,29 +941,70 @@ public function install( $args, $assoc_args ) {
...
+			'requires_wp'      => 'RequiresWP',
+			'requires_php'     => 'RequiresPHP',
+			'requires_plugins' => 'RequiresPlugins',
+		];
+
+		foreach ( $require_fields as $field_key => $data_key ) {
+			if ( isset( $plugin_data[ $data_key ] ) && ! empty( $plugin_data[ $data_key ] ) ) {
+				$plugin_obj->{$field_key} = $plugin_data[ $data_key ];
+			}
+		}
```
swissspidy, line 1004: Can we simplify this?

```suggestion
			'name'             => Utils\get_plugin_name( $file ),
			'title'            => $plugin_data['Name'],
			'author'           => $plugin_data['Author'],
			'version'          => $plugin_data['Version'],
			'description'      => wordwrap( $plugin_data['Description'] ),
			'status'           => $this->get_status( $file ),
			'requires_wp'      =>  ! empty( $plugin_data['requires_wp'] ) ? $plugin_data['requires_wp'] : '',
			'requires_php'     => ! empty( $plugin_data['requires_php'] ) ? $plugin_data['requires_php'] : '',
			'requires_plugins' => ! empty( $plugin_data['requires_plugins'] ) ? $plugin_data['requires_plugins'] : '',
		];
```

swissspidy:
Now that #417 is merged, merging `main` into this branch should fix the tests

swissspidy (approved):

PR #417, opened 2024-04-17
Update the theme in feature tests

tfirdaus:
This Pull Request updates the theme to install in the feature tests from `p2` to `twentytwelve` since the `p2` has been removed from WP.org repository.

### Reference 

- https://github.com/wp-cli/wp-cli-tests/issues/200
- https://themes.trac.wordpress.org/ticket/30322

### DIFF
--- a/features/theme-delete.feature
+++ b/features/theme-delete.feature
@@ -2,32 +2,32 @@ Feature: Delete WordPress themes
 
   Background:
     Given a WP install
-    And I run `wp theme install twentytwelve`
+    And I run `wp theme install twentytwelve --force`
 
   Scenario: Delete an installed theme
-    When I run `wp theme delete p2`
+    When I run `wp theme delete twentytwelve`
     Then STDOUT should be:
       """
-      Deleted 'p2' theme.
+      Deleted 'twentytwelve' theme.
       Success: Deleted 1 of 1 themes.
       """
     And the return code should be 0
 
   Scenario: Delete an active theme
-    When I run `wp theme activate p2`
+    When I run `wp theme activate twentytwelve`
     Then STDOUT should not be empty
 
-    When I try `wp theme delete p2`
+    When I try `wp theme delete twentytwelve`
     Then STDERR should be:
       """
-      Warning: Can't delete the currently active theme: p2
+      Warning: Can't delete the currently active theme: twentytwelve
       Error: No themes deleted.
       """
 
-    When I try `wp theme delete p2 --force`
+    When I try `wp theme delete twentytwelve --force`
     Then STDOUT should contain:
       """
-      Deleted 'p2' theme.
+      Deleted 'twentytwelve' theme.
       """
 
   Scenario: Delete all installed themes
@@ -108,7 +108,7 @@ Feature: Delete WordPress themes
     Then STDOUT should be empty
 
   Scenario: Attempting to delete a theme that doesn't exist
-    When I run `wp theme delete p2`
+    When I run `wp theme delete twentytwelve`
     Then STDOUT should not be empty
 
     When I try the previous command again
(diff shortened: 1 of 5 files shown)


ernilambar:
You can try with replacing 
```
And I run `wp theme install p2`
```
with 
```
And I try `wp theme install twentytwelve --force`
```

This will install theme even if it already exists. With this we will be sure that TwentyTwelve theme is available in any case.

tfirdaus:
I tried using the `--force` flag, but it seems like it's going to require more updates to the initial feature spec than the necessary. Here's an example where it didn't work as expected. Since the `twentytwenty` theme is already installed, if I use `--force`, I'll have to change the output message from **"Theme installed successfully."** to **"Theme updated successfully."**

Screenshot 2024-04-24 at 1 49 11 PM

@ernilambar @swissspidy Thought on this. Do you think this is the preferable approach? 
Ideally I'd prefer to keep the updates to a minimum.

ernilambar:
I think you added `--force` to the command in `When`. We should not be updating those arguments. You can add **And I run `wp theme delete --all --force`** in Given (after line 4) in  `theme-install.feature`

features/theme-delete.feature
```diff
@@ -2,7 +2,7 @@ Feature: Delete WordPress themes
 
   Background:
     Given a WP install
-    And I run `wp theme install p2`
+    And I run `wp theme install twentytwelve`
```
swissspidy, line 5: I think that's what @ernilambar meant:

```suggestion
    And I run `wp theme delete --all --force`
    And I run `wp theme install twentytwelve`
```

features/theme-delete.feature
```diff
@@ -2,7 +2,7 @@ Feature: Delete WordPress themes
...
 
   Scenario: Delete an installed theme
     When I run `wp theme delete p2`
```
swissspidy, line 8:
```suggestion
    When I run `wp theme delete twentytwelve`
```

Because we're now installing twentytwelve in `Background`

features/theme.feature
```diff
@@ -3,7 +3,7 @@ Feature: Manage WordPress themes
   Scenario: Installing and deleting theme
     Given a WP install
 
-    When I run `wp theme install p2`
+    When I run `wp theme install twentytwelve`
     Then STDOUT should not be empty
 
     When I run `wp theme status p2`
```
swissspidy, line 9: The following steps all need to be updated as well because we're no longer installing p2

features/theme.feature
```diff
@@ -74,7 +74,7 @@ Feature: Manage WordPress themes
   Scenario: Install a theme, activate, then force install an older version of the theme
     Given a WP install
 
-    When I run `wp theme install p2 --version=1.4.2`
+    When I run `wp theme install twentytwelve --version=1.4.2`
     Then STDOUT should not be empty
 
     When I run `wp theme list --name=p2 --field=update_version`
```
swissspidy, line 80: The following steps all need to be updated as well because we're no longer installing p2

features/theme-delete.feature
tfirdaus: Thank you all for clarifying it :)

swissspidy:
There are still a lot of tests referencing `p2`. I'd probably start by replacing all of those with `twentytwelve`

tfirdaus:
@ernilambar this error is a bit weird to me. Somehow it does not happen when I run this test in local. What do you think I did wrong here?

```
009 Scenario: Parent theme is active when its child is active # features/theme.feature:577
      And I run `wp theme delete twentytwelve --force`        # features/theme.feature:579
        $ wp theme delete twentytwelve --force
        Success: Theme already deleted.
```

ernilambar (approved):

tfirdaus:
Thanks @ernilambar for your help on this PR 👏

PR #421, opened 2024-04-26
Add support for GitHub release installation

tfirdaus:
Related to #91 and resuming #386.

PR #423, opened 2024-04-30
Fix visibility in command class methods

ernilambar:
Introduced in https://github.com/wp-cli/extension-command/pull/421

* Change method visibility from `public` to `private`

When kept `public`, `cmd_dump()` was taking those as commands which should not be. https://github.com/ernilambar/wpcli/blob/main/src/manifest.json#L4442

### DIFF
--- a/src/WP_CLI/CommandWithUpgrade.php
+++ b/src/WP_CLI/CommandWithUpgrade.php
@@ -825,7 +825,7 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
 	 *
 	 * @return array{ name: string, url: string }|\WP_Error
 	 */
-	public function get_the_latest_github_version( $repo_slug ) {
+	protected function get_the_latest_github_version( $repo_slug ) {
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -825,7 +825,7 @@ private function parse_url_host_component( $url, $component ) {
 	 *
 	 * @return array{ name: string, url: string }|\WP_Error
 	 */
-	public function get_the_latest_github_version( $repo_slug ) {
+	private function get_the_latest_github_version( $repo_slug ) {
```
thelovekesh, line 828: These methods were public by default so they could be used somewhere else. So it would be good to set the visibility to `protected` instead of `private`.

Something like this has broken things in the past - https://github.com/wp-cli/wp-cli-tests/pull/170

thelovekesh (changes requested):

thelovekesh (approved):

PR #426, opened 2024-05-07
Add `--force-check` flag to `wp plugin list` and `wp theme list`.

baizmandesign:
Hi @danielbachhuber,

In the process of updating my fork, I somehow accidentally closed [the existing pull request](https://github.com/wp-cli/extension-command/pull/399) for this feature.

Here's the code and tests. Please let me know if you have any feedback!

Saul

### DIFF
--- a/features/plugin-list.feature
+++ b/features/plugin-list.feature
@@ -1,71 +1,56 @@
 Feature: List WordPress plugins
 
-  Background:
+  Scenario: Refresh update_plugins transient when listing plugins with --force-check flag
     Given a WP install
     And I run `wp plugin uninstall --all`
     And I run `wp plugin install hello-dolly`
-    And a setup.php file:
+    And a update-transient.php file:
       """
       <?php
-      $plugin_transient = get_site_transient('update_plugins') ;
-      # manually set the version
-      $fake_version = '10.0.0';
-      $plugin_transient->no_update["hello-dolly/hello.php"]->new_version = $fake_version;
-      $return = set_site_transient( 'update_plugins', $plugin_transient );
+      $transient = get_site_transient( 'update_plugins' );
+      $transient->response['hello-dolly/hello.php'] = (object) array(
+        'id'          => 'w.org/plugins/hello-dolly',
+        'slug'        => 'hello-dolly',
+        'plugin'      => 'hello-dolly/hello.php',
+        'new_version' => '100.0.0',
+        'url'         => 'https://wordpress.org/plugins/hello-dolly/',
+        'package'     => 'https://downloads.wordpress.org/plugin/hello-dolly.100.0.0.zip',
+      );
+      $transient->checked = array(
+        'hello-dolly/hello.php' => '1.7.2',
+      );
+      unset( $transient->no_update['hello-dolly/hello.php'] );
+      set_site_transient( 'update_plugins', $transient );
+      WP_CLI::success( 'Transient updated.' );
       """
-#    And a wp-content/mu-plugins/test-plugin-update.php file:
-#      """
-#      <?php
-#      /**
-#       * Plugin Name: Test Plugin Update
-#       * Description: Fakes installed plugin's data to verify plugin version mismatch
-#       * Author: WP-CLI tests
-#       */
-#
-#      add_filter( 'site_transient_update_plugins', function( $value ) {
-#          if ( ! is_object( $value ) ) {
-#              return $value;
-#          }
-#          $fake_version = '10.0.0';
-#
-#          unset( $value->response['hello-dolly/hello.php'] );
-#          $value->no_update['hello-dolly/hello.php']->new_version = $fake_version;
-#
-#          return $value;
-#      } );
-#      ?>
-#      """
 
-  Scenario: Refresh update_plugins transient when listing plugins with --force-check flag
+    # Populates the initial transient in the database
+    When I run `wp plugin list --fields=name,status,update`
+    Then STDOUT should be a table containing rows:
+      | name         | status   | update |
+      | hello-dolly  | inactive | none   |
 
-    # Listing the plugins will populate the transient in the database
-    When I run `wp plugin list`
-    Then STDOUT should not be empty
-    And save STDOUT as {PLUGIN_LIST_ORIGINAL_VERSIONS}
+    # Modify the transient in the database to simulate an update
+    When I run `wp eval-file update-transient.php`
+    Then STDOUT should be:
+      """
+      Success: Transient updated.
+      """
 
-    # Write a test value
-    When I run `wp eval-file setup.php`
-    Then STDOUT should be empty
+    # Verify the fake transient value produces the expected output
+    When I run `wp plugin list --fields=name,status,update`
+    Then STDOUT should be a table containing rows:
+      | name         | status   | update    |
+      | hello-dolly  | inactive | available |
 
-    # Get value of plugin transient
-    #When I run `wp option get _site_transient_update_plugins`
-    #Then STDOUT should be empty
+    # Repeating the same command again should produce the same results
+    When I run `wp plugin list --fields=name,status,update`
+    Then STDOUT should be a table containing rows:
+      | name         | status   | update    |
+      | hello-dolly  | inactive | available |
 
-    # Run list again
-    When I run `wp plugin list`
-    Then STDOUT should be empty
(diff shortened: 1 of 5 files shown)


ernilambar:
@baizmandesign Please make sure all tests except PHP 8.4 are passing in Actions.

features/theme-list.feature
```diff
@@ -0,0 +1,64 @@
...
+
+  Scenario: Refresh update_themes transient when listing themes with --force-check flag
+    Given a WP install
+    And I run `wp theme delete --all --force`
+    And I run `wp theme install --force twentytwentyfour --version=1.1`
+    And a update-transient.php file:
+      """
+      <?php
+      $transient = get_site_transient( 'update_themes' );
+      $transient->response['twentytwentyfour'] = (object) array(
```
ernilambar, line 11: Using TwentyTwentyFour theme in feature test will give error in lower PHP versions. So please use may be TwentyTwelve.

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -532,6 +532,16 @@ static function ( $result ) {
 	// phpcs:ignore PSR2.Methods.MethodDeclaration.Underscore -- Whitelisting to provide backward compatibility to classes possibly extending this class.
 	protected function _list( $_, $assoc_args ) {
 
+		// If `--force-check` and `--skip-update-check` flags are both present, abort.
+		if ( true === (bool) Utils\get_flag_value( $assoc_args, 'force-check', false ) and true === (bool) Utils\get_flag_value( $assoc_args, 'skip-update-check', false ) ) {
```
ernilambar, line 536: PHPCS error: Logical operator "and" is prohibited; use "&&" instead.

features/plugin-list.feature
```diff
@@ -0,0 +1,64 @@
...
+      Success: Transient updated.
+      """
+
+    # Verify the fake transient value produces the expected output
+    When I run `wp plugin list --fields=name,status,update`
+    Then STDOUT should be a table containing rows:
+      | name         | status   | update    |
+      | hello-dolly  | inactive | available |
+
+    # Repeating the same command again should produce the same results
```
ernilambar, line 46: WPCLI has `I run the previous command again` step also. You can trying using this here.

ernilambar: Other spacing issues are also there. You can use `composer run phpcbf` to fix such issues.

baizmandesign: Thanks for all of your feedback, @ernilambar. I'll make the changes and commit the revisions shortly.

baizmandesign:
Hi, @ernilambar. I added another commit, and it failed three tests: 2 related to PHP 8.4 and one test due to a GitHub API rate limit issue. Is there a way to re-run the one failed test unrelated to PHP 8.4?

swissspidy:
PHP 8.4 tests are expected to fail, PHP 8.4 is still in alpha and WP-CLI is not yet compatible. Feel free to ignore those.

baizmandesign:
Thanks for the info, @swissspidy. Is there a way to re-run the one test unrelated to PHP 8.4 that failed?

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -532,6 +532,16 @@ static function ( $result ) {
 	// phpcs:ignore PSR2.Methods.MethodDeclaration.Underscore -- Whitelisting to provide backward compatibility to classes possibly extending this class.
 	protected function _list( $_, $assoc_args ) {
 
+		// If `--force-check` and `--skip-update-check` flags are both present, abort.
+		if ( true === (bool) Utils\get_flag_value( $assoc_args, 'force-check', false ) && true === (bool) Utils\get_flag_value( $assoc_args, 'skip-update-check', false ) ) {
+			WP_CLI::error( "{$this->item_type} updates cannot be both force-checked and skipped. Choose one." );
```
ernilambar, line 537: We generally keep messages starting with capital letter. You can use `ucfirst()` here. Other than this PR looks ok to me.

baizmandesign:
Hi, @ernilambar. I've made the changes you requested. Please let me know if there's anything else I need to do. Thanks!

BrianHenryIE:
Looks good to me; tests run locally.

I like the `When I run wp eval-file update-transient.php` trick.

mrsdizzie:
I notice that we use `--force` in most other commands for similar behavior. Would it be better to name this `--force` and not `--force-check` to match existing commands?

swissspidy:
It's a good question. For commands like `install` or `update`, the intention of `--force` seems clear.

But for `wp plugin/theme list`, only `--force` seems too ambiguous to me.

BrianHenryIE:
`--force` makes sense in terms of something being written which is already known. It's more of a push word, whereas the point of this PR is to pull/fetch the latest information. I found myself using `--refresh` in a premium plugin licence update command I was writing. 

I do see `--skip-check` elsewhere, so `--force-check` makes sense as its opposite. Also worth considering that `--flag` automatically comes with `--no-flag` ([utils.php](https://github.com/wp-cli/wp-cli/blob/a339dca576df73c31af4b4d8054efc2dab9a0685/php/utils.php#L983)) ([Configurator.php](https://github.com/wp-cli/wp-cli/blob/a339dca576df73c31af4b4d8054efc2dab9a0685/php/WP_CLI/Configurator.php#L172)).

`--force-check` is probably fine here.

In terms of premium plugins, this does now suggest that when the `update_plugins` transient value does not have the premium plugin's update status, it should fetch it from its remote server and _not_ to use cached information (which is fine, just not so explicit before).

mrsdizzie:
Tempting to just remove the flag and make this the default behavior if we're also having the discussion of a 3.0 release that can include 'breaking' changes https://github.com/wp-cli/wp-cli/issues/6026

I know personally I always want this command to do a live check if I'm running it directly from the command line and have been bitten in the past looking for a security update that doesn't show because it wasn't there the last time wp itself checked

mrsdizzie:
> Having both force-check and skip-update-check is certainly confusing 🤔

I agree it feels weird to have both a `--force-check` and `--skip-update-check` (I don't really understand the benefit of `--skip-update-check` but I know we can't remove it now). It implies the state of updates is really just an unknown which doesn't seem right for this tool. This is another reason to have it just always check for updates (imo), and the current flag could prevent that if wanted.

I understand why WordPress core shouldn't do an api call for every visit to the plugins page, but I'm not sure wp cli needs that same restriction (and I think most people using wp cli would prefer the --force behavior).

If we don't want to make a force check the default behavior I think we just need to accept this new flag as awkward but merge this PR. Personally I've gotten used to the more awkward `wp transient delete --all; wp plugin list` so it would be an improvement on that :)

baizmandesign:
@mrsdizzie I'm the author of this PR, and I would *love* the default behavior of WP CLI to be to always check for the latest versions of plugins and themes and that this `--force-check` flag wasn't necessary at all. It's confusing (and frustrating) that running the list commands only occasionally checks for the newest versions of plugins and themes—whose behavior is unstated and whose "occasional" frequency is unknown. I do not recall how I eventually learned that to retrieve the most up-to-date plugin and theme versions, I had to _delete transients_, but I recall this step was not obvious.

Adding this functionality as the default behavior would not be a breaking change (I don't think), and I suspect it would prevent a small amount of confusion for a subset of WP CLI users.

swissspidy:
WP-CLI has typically avoided changing default behavior when it comes to cases like this with transients/cache. So in many cases the user needs to manually do that. (see also [this related discussion](https://github.com/wp-cli/ideas/issues/179))
This situation here is similar to `wp core check-updates` , where `wp_version_check()` defaults to a transient unless forced. But `wp_update_plugins()` and `wp_update_themes(`) don't have such a flag. One needs to manually delete the transients. These commands have `skip-update-check` though, but that doesn't make sense when omitting it doesn't actually force the check. I would consider that a bug.

So to me it makes sense to change the default behavior to force (i.e. delete the transient) unless `skip-update-check` is used.

mrsdizzie:
Could probably just put a `delete_site_transient( $this->upgrade_transient );` inside here above the `call_user_func`
https://github.com/wp-cli/extension-command/blob/33afe4f86812a3173c20a584c6a070e68dfafc4c/src/WP_CLI/CommandWithUpgrade.php#L544-L547

Interesting that comment suggests to me that this should already be happening

swissspidy:
Yeah that's where I would put it 👍

PR #436, opened 2024-12-06
Check WordPress and PHP requirements before installing a theme or plugin

mrsdizzie:
This matches the behavior of WordPress core which will refuse to install a plugin if the local copy of WordPress or PHP don't meet the minimum requirements for the most recent version as listed by the plugin authors.

Unfortunately the api is limited and only provides these requirement details for the most recent version, so it isn't possible to find an older version that might work. As a compromise, this code doesn't check requirements if a user provides a specific `--version` since we can't know the requirements for anything other than the latest version and assume if somebody specifies a version they know it will work or want to try anyway.

### DIFF
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -238,6 +238,7 @@ Feature: Install WordPress plugins
     Success: Plugin already installed.
     """
 
+  @require-php-7
   Scenario: Can't install plugin that requires a newer version of WordPress
     Given a WP install
 
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -237,3 +237,35 @@ Feature: Install WordPress plugins
     Plugin 'site-secrets' activated.
     Success: Plugin already installed.
     """
+
+  Scenario: Can't install plugin that requires a newer version of WordPress
+    Given a WP install
+
+    When I run `wp core download --version=6.4 --force`
+    And I run `rm -r wp-content/themes/*`
+
+    And I try `wp plugin install wp-super-cache`
+    Then STDERR should contain:
+    """
+    Warning: wp-super-cache: This plugin does not work with your version of WordPress
+    """
+
+    And STDERR should contain:
+    """
+    Error: No plugins installed.
+    """
+
+  @less-than-php-7.4
+  Scenario: Can't install plugin that requires a newer version of PHP
+    Given a WP install
+
+    And I try `wp plugin install contact-form-7`
+    Then STDERR should contain:
+    """
+    Warning: contact-form-7: This plugin does not work with your version of PHP
+    """
+
+    And STDERR should contain:
+    """
+    Error: No plugins installed.
+    """
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -255,7 +255,7 @@ Feature: Install WordPress plugins
     Error: No plugins installed.
     """
 
-  @less-than-php-7.4
+  @less-than-php-7.4 @require-wp-6.6
   Scenario: Can't install plugin that requires a newer version of PHP
     Given a WP install
 
--- a/features/plugin-install.feature
+++ b/features/plugin-install.feature
@@ -171,7 +171,7 @@ Feature: Install WordPress plugins
     When I run `rm wp-content/plugins/akismet/akismet.php`
     Then the return code should be 0
 
-    When I try `wp plugin install akismet`
+    When I try `wp plugin install akismet --ignore-requirements`
     Then STDERR should contain:
       """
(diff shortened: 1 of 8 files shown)


mrsdizzie:
For reference, this is what it looks like trying to install a plugin that doesn't meet requirements in the WordPress dashboard:

Screenshot 2024-12-06 at 13 22 22

mrsdizzie:
Interesting, this now reveals issues with existing tests like:
```
Warning: akismet: This plugin does not work with your version of WordPress. Minimum WordPress requirement is 5.8
```
When running the WordPress 3.7 test. Several similar errors in the WordPress 6.2 test as many of the plugins we use for the tests actually require newer versions of WordPress.

If the PR seems OK otherwise I can fix those, either by finding new plugins or just by specifying a specific plugin version for those tests (which bypasses this new check).

src/Plugin_Command.php
```diff
@@ -590,6 +591,20 @@ protected function install_from_repo( $slug, $assoc_args ) {
 
 		if ( isset( $assoc_args['version'] ) ) {
 			self::alter_api_response( $api, $assoc_args['version'] );
+		} else {
+			$requires_php = isset( $api->requires_php ) ? $api->requires_php : null;
+			$requires_wp  = isset( $api->requires ) ? $api->requires : null;
+
+			$compatible_php = empty( $requires_php ) || version_compare( PHP_VERSION, $requires_php, '>=' );
+			$compatible_wp  = empty( $requires_wp ) || version_compare( $wp_version, $requires_wp, '>=' );
```
swissspidy, line 599: Should we mimic `is_wp_version_compatible()` and strip off any suffixes and turn versions like `6.8.0` into `6.8`?

mrsdizzie: Good point I think we should probably do that so that version_compare always works properly. I think to get a minimal mimic of `is_wp_version_compatible()` We could do this:
```
list($core_version) = explode('-', $wp_version);
$core_version = implode('.', array_slice(explode('.', $core_version), 0, 2));
```

features/theme-install.feature
```diff
@@ -125,3 +125,35 @@ Feature: Install WordPress themes
...
+    """
+    Warning: twentytwentyfive: This theme does not work with your version of WordPress.
+    """
+
+    And STDERR should contain:
+    """
+    Error: No themes installed.
+    """
+
+  @less-than-php-7.4 @require-wp-5.6
```
mrsdizzie, line 146: I think these PHP version related tests are always going to be flaky. It is hard to find something that has a higher PHP requirement and doesn't also have a high WP requirement (meaning the test can't run on older versions). Also I suspect these requirements are going to change over time, where an upstream change can break it.

Outside of actually uploading our own test plugin to wp.org where we can artificially  set requirements for this test, I wonder if it just better not to have a test for this in favor of future stability.

features/plugin-install.feature
```diff
@@ -237,3 +237,35 @@ Feature: Install WordPress plugins
     Plugin 'site-secrets' activated.
     Success: Plugin already installed.
     """
+
+  Scenario: Can't install plugin that requires a newer version of WordPress
```
swissspidy, line 241:
```suggestion
  @require-php-7
  Scenario: Can't install plugin that requires a newer version of WordPress
```

features/theme-install.feature
```diff
@@ -125,3 +125,35 @@ Feature: Install WordPress themes
       """
       twentyeleven
       """
+
+  Scenario: Can't install theme that requires a newer version of WordPress
```
swissspidy, line 129:
```suggestion
  @require-php-7
  Scenario: Can't install theme that requires a newer version of WordPress
```

features/theme-install.feature
swissspidy: It would be easier if the API had requirements info even for older versions, but alas.

Let's keep it for now and see how flaky it will actually become.

swissspidy:
Should we perhaps add an `--ignore-requirements` arg? As long as you don't activate the plugin/theme, installing an incompatible version is not a problem.

This will also be relevant for #407 (plugin dependencies).

mrsdizzie:
> Should we perhaps add an `--ignore-requirements` arg? As long as you don't activate the plugin/theme, installing an incompatible version is not a problem.
> 
> This will also be relevant for #407 (plugin dependencies).

Sounds good, I added that arg. The only thing left is what to do about failing tests on WP 3.7 and WP 6.2. They are all for the same reason that the plugins we use for these tests have a higher WP requirement and now `wp plugin/theme install` fails with an appropriate error.

swissspidy:
The tests involving `wporg_last_updated` are really flaky right now, so I'm ignoring those failures for now.

Issue #437, opened 2024-12-09
`plugin_deleted` hook is not firing in WordPress when `wp plugin uninstall` is called in wp cli


dkoston:
## Bug Report

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [X] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

When calling `wp plugin uninstall $plugin_name` the `deleted_plugin` event does not fire in WordPress

Context, we have a plugin which tracks plugin related events.

When calling `wp plugin deactivate`, we see this hook fire:

`add_action('deactivated_plugin', array(&$this, 'deactivated_plugin_log'));`

When calling `wp plugin uninstall`, this hook does not fire:

`add_action('deleted_plugin', array($this, 'deleted_plugin_log'));`

However, when deleting the plugin in WP Admin, the hook fires.

**Describe how other contributors can replicate this bug**

- Install any plugin in WordPress
- Add a plugin with the above hooks
- click in WP Admin to delete the plugin and see the hook has fired (you'll need some logic to log, send http request, etc in your hook function...let me know if you want me to provide that logic
- install any plugin
- call `wp plugin uninstall $plugin` and see that the hook has not fired

> NOTE: also tried `wp plugin delete` but since that technically doesn't do the uninstall / deactivate, I wouldn't expect that to call the hook. `wp plugin deactivate` does fire the `deactivated_plugin` hook so it appears that `wp plugin uninstall` is using a non-standard way to delete a plugin?

**Describe what you would expect as the correct outcome**

WP CLI should use the same functions in WP Core that WP-Admin does so that the same hooks are activated

**Let us know what environment you are running this on**

```
OS:	Linux 4.18.0-513.11.1.lve.el8.x86_64 #1 SMP Thu Jan 18 16:21:02 UTC 2024 x86_64
Shell:	/usr/local/cpanel/bin/noshell
PHP binary:	/opt/alt/php80/usr/bin/php
PHP version:	8.0.30
php.ini used:	/opt/alt/php80/etc/php.ini
MySQL binary:	/bin/mysql
MySQL version:	mysql  Ver 8.0.37 for Linux on x86_64 (MySQL Community Server - GPL)
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
[...]

PR #438, opened 2024-12-10
Plugin_Command: Add delete_pugin and deleted_plugin hooks to delete_plugin(). Remove .10n.php files when removing language files. Do not count failed delete_plugins as successes, remove deleted plugins from plugin update list

dkoston:
Fixes: https://github.com/wp-cli/extension-command/issues/437

1. Add missing hooks for `delete_plugin` and `deleted_plugin` to `delete_plugin()`
2. Add function comment to `delete_plugin()`
3. Also remove .10.n.php files when removing language files as per https://github.com/WordPress/WordPress/blob/e8b5d1a702e9fbf6a1f225df6501ea2e517938ec/wp-admin/includes/plugin.php#L1022
4. Add additional logic branch for a plugin which was uninstalled successfully but not deleted.
5. Remove any successfully deleted plugins from plugin update list.

Excuse my lack of tests and general ignorance of how to code on the WP CLI codebase. This is my first PR and I'm not seeing any testing like I would normally do for production projects:

1. I'd expect a set of functional tests which would test conditional logic branches here (ensure that --skip-delete doesn't actually call $this->delete_plugin by using a spy)
2. Typically I'd unit test and create smaller functions from the logic here as well and since these functions are essentially all calling WordPress functions, I'd do this:
  a. Would update WordPress core to take any logic that's being copy/pasted into WP CLI and extract it into functions that can be called, rather than duplicating the logic as it would mean only 1 set of functions to maintain.
 b. Would be create a set of table-driven tests to test all the logic branches of `uninstall()` which have spies to make sure that the appropriate WordPress core functions are called (or not) depending on the flags passed 

However, I don't know the logic and history behind why this codebase has duplicate logic. I'm not sure if there are stability concerns relying on core directly like that or other reason.

Let me know if there's ways I can improve the contribution and what levels of testing I can apply to make sure we have confidence in the current changes and to prevent regressions.

Thanks!


src/Plugin_Command.php
```diff
@@ -1148,13 +1152,35 @@ public function uninstall( $args, $assoc_args = array() ) {
 				}
 			}
 
-			if ( ! Utils\get_flag_value( $assoc_args, 'skip-delete' ) && $this->delete_plugin( $plugin ) ) {
-				WP_CLI::log( "Uninstalled and deleted '$plugin->name' plugin." );
+			if ( ! Utils\get_flag_value( $assoc_args, 'skip-delete' ) ) {
+				if ( $this->delete_plugin( $plugin ) ) {
+					$deleted_plugin_files[] = $plugin->file;
+					WP_CLI::log( "Uninstalled and deleted '$plugin->name' plugin." );
```
dkoston, line 1158: This logic branch didn't appear to be covered before (uninstall success and delete failure). I don't know how to reproduce it using the purely spec driven tests as it appears to need a mock

src/Plugin_Command.php
```diff
@@ -1148,13 +1152,35 @@ public function uninstall( $args, $assoc_args = array() ) {
...
+					++$errors;
+					continue;
+				}
 			} else {
 				WP_CLI::log( "Ran uninstall procedure for '$plugin->name' plugin without deleting." );
 			}
 			++$successes;
 		}
+
+		// Remove deleted plugins from the plugin updates list.
```
dkoston, line 1171: the assumption here is that only deleted plugins should be removed as the others still have files around? And as a result it would be good to keep those files up-to-date

dkoston:
Not sure what to do about the hook name linter which is not compatible with the legacy WordPress hook names:

```
✗ vendor/bin/phpcs -q --report=full --report-checkstyle=/tmp/phpcs-checkstyle-report.xml

FILE: src/Plugin_Command.php
-------------------------------------------------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
-------------------------------------------------------------------------------------------------------------------------
 1510 | ERROR | Hook names invoked by a theme/plugin should start with the theme/plugin prefix. Found: "delete_plugin".
 1534 | ERROR | Hook names invoked by a theme/plugin should start with the theme/plugin prefix. Found: "deleted_plugin".
-------------------------------------------------------------------------------------------------------------------------
```

src/Plugin_Command.php
```diff
@@ -1474,7 +1500,15 @@ private function get_details( $file ) {
...
 
+	/**
+	 * Performs deletion of plugin files
+	 *
+	 * @param $plugin - Plugin fetcher object (name, file)
+	 * @return bool - If plugin was deleted
+	 */
 	private function delete_plugin( $plugin ) {
+		do_action( 'delete_plugin', $plugin->file );
+
```
dkoston, line 1511: Didn't want to `// phpcs:ignore ` as that would remove all checks. It's not clear to me how I would understand which custom matchers these match to as they have "names" but nothing matches names or comments in https://github.com/wp-cli/extension-command/blob/main/phpcs.xml.dist:

```
FILE: src/Plugin_Command.php
-------------------------------------------------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
-------------------------------------------------------------------------------------------------------------------------
 1510 | ERROR | Hook names invoked by a theme/plugin should start with the theme/plugin prefix. Found: "delete_plugin".
 1534 | ERROR | Hook names invoked by a theme/plugin should start with the theme/plugin prefix. Found: "deleted_plugin".
-------------------------------------------------------------------------------------------------------------------------
```

dkoston: Ah, nm, found it: https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php

dkoston: maybe...still not sure why phpcs doesn't just provide the class name of the sniff that failed

dkoston: 🤷 https://github.com/search?q=repo%3AWordPress%2FWordPress-Coding-Standards+%22Hook+names+invoked+by+a+%22&type=code

swissspidy: The error you wanna ignore is `WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound`

I isually use `vendor/bin/phpcs --report=xml` to find that out quickly.

dkoston: Ah, turns out you need to use `-s` to see the source of the phpcs rules

swissspidy:
Thanks for your PR, it's much appreciated!

> Excuse my lack of tests and general ignorance of how to code on the WP CLI codebase. This is my first PR and I'm not seeing any testing like I would normally do for production projects:

WP-CLI uses [Behat](https://behat.org/) as its functional test suite. Functional tests are different than unit tests in that they execute the entire WP-CLI command, and ensure they always work as expected. Every repository has a `features/` directory with one or more [YAML](https://yaml.org/)-formatted `*.feature` files. Here, `plugin-delete.feature` and `plugin-uninstall.feature` are relevant.

See https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-and-writing-tests and https://www.youtube.com/playlist?list=PL_B8Y6K6MH2d6T7pYa6dloUgk67mfBT4K

One example test would be to install a plugin, install its translations, then run `wp plugin delete` and verify that the translations were properly deleted.

dkoston:
> Thanks for your PR, it's much appreciated!
> 
> > Excuse my lack of tests and general ignorance of how to code on the WP CLI codebase. This is my first PR and I'm not seeing any testing like I would normally do for production projects:
> 
> WP-CLI uses [Behat](https://behat.org/) as its functional test suite. Functional tests are different than unit tests in that they execute the entire WP-CLI command, and ensure they always work as expected. Every repository has a `features/` directory with one or more [YAML](https://yaml.org/)-formatted `*.feature` files. Here, `plugin-delete.feature` and `plugin-uninstall.feature` are relevant.
> 
> See https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-and-writing-tests and https://www.youtube.com/playlist?list=PL_B8Y6K6MH2d6T7pYa6dloUgk67mfBT4K
> 
> One example test would be to install a plugin, install its translations, then run `wp plugin delete` and verify that the translations were properly deleted.

Thanks @swissspidy 

Sorry, I think I was unintentionally using language around testing framework / style vs discussing what is actually tested. BDD is a totally fine approach.


When I look at: https://github.com/wp-cli/extension-command/blob/main/features/plugin-uninstall.feature

The only thing that checks something other than STDERR/STDOUT and the exit code is this test:

https://github.com/wp-cli/extension-command/blob/main/features/plugin-uninstall.feature#L94-L114

There's no actual validation that any plugins have been installed, removed, what logic branches have been exercised, etc.

Its definitely fine to take a functional style and agreed that one should focus on testing the interfaces that people interact with but what I was missing is a way to understand how to exercise the various logic branches in the changes I've made and to validate that things are "actually" happening (other than the files being present or not) because IMO, the presence of text on STDOUT/STDERR doesn't indicate anything other than a logic branch with the `WPCLI::log()` function was executed.

Not trying to be contrarian, just trying to make sure that I can be confident in the changes that I'm bringing to the project and help future contributors from not having to manually run WP CLI with all permutations to validate success.

I'll dig more into Behat to see if there are ways to validate the underlying logic branches and confirm that they are being exercised.

features/plugin-uninstall.feature
```diff
@@ -11,6 +11,39 @@ Feature: Uninstall a WordPress plugin
       Success: Uninstalled 1 of 1 plugins.
       """
     And the return code should be 0
+    And STDERR should be empty
+    And the wp-content/plugins/akismet directory should not exist
+
+
+  Scenario: Uninstall an installed plugin but do not delete its files
```
dkoston, line 18: added a test for missing `--skip-delete` logic branch

features/plugin-uninstall.feature
```diff
@@ -11,6 +11,39 @@ Feature: Uninstall a WordPress plugin
...
+    Then STDOUT should be:
+      """
+      Ran uninstall procedure for 'akismet' plugin without deleting.
+      Success: Uninstalled 1 of 1 plugins.
+      """
+    And the return code should be 0
+    And STDERR should be empty
+    And the wp-content/plugins/akismet directory should exist
+
+  Scenario: Uninstall a plugin that is not in a folder and has custom name
```
dkoston, line 29: added test for plugins which are php files

features/plugin-uninstall.feature
```diff
@@ -112,3 +164,5 @@ Feature: Uninstall a WordPress plugin
       """
     And the wp-content/languages/plugins/wordpress-importer-fr_FR.mo file should not exist
     And the wp-content/languages/plugins/wordpress-importer-fr_FR.po file should not exist
+    And the wp-content/languages/plugins/wordpress-importer-fr_FR.l10n.php file should not exist
```
dkoston, line 167: @swissspidy I added the test here for the php language file but its not clear how to me that I can test these things:

1. Test that if I have the `update_plugins` transient that any deleted plugins are removed
2. Test that if uninstall succeeds and delete fails that the user will get the error for that

I think those are the remaining logic branches to be tested but I don't see any code coverage reporting with Behar

dkoston:
Seeing failed tests outside the scope of the PR:

https://github.com/wp-cli/extension-command/blob/main/features/plugin-list-wporg-status.feature#L29

```
001 Scenario: Install plugins and check the status on wp.org. # features/plugin-list-wporg-status.feature:4
      Then STDOUT should be a table containing rows:          # features/plugin-list-wporg-status.feature:30
        name	wporg_last_updated
        akismet	2024-12-09
        hello	
        no-longer-in-directory	2017-11-13
        never-wporg	
        wordpress-importer	
        no-mail	
        polyfills	
         (Exception)
```

This seems odd as the test above has a different set of outputs for `wp plugin list`:

```
    When I run `wp plugin list --fields=name,wporg_status`
      | name                   | wporg_status    |
      | wordpress-importer     | active          |
      | no-longer-in-directory | closed          |
      | never-wporg            |                 |
```


Not sure how to proceed

features/plugin-uninstall.feature
swissspidy: > Test that if I have the update_plugins transient that any deleted plugins are removed from it

You can run `wp transient get update_plugins` in the test and verify that the plugin is not part of the output.

> Test that if uninstall succeeds and delete fails that the user will get the error for that

You could for example check STDERR/STDOUT and whether the files still exist.

swissspidy:
> There's no actual validation that any plugins have been installed, removed, what logic branches have been exercised, etc.
> to validate that things are "actually" happening

You can also use other commands to verify that a plugin is no longer present, e.g. by verifying whether `wp plugin list` includes the plugin, or if the directory no longer exists.

> I didn't see any tooling to run tests locally. Really sorry about using all these compute cycles on GitHub Actions to get these tests passing. Please let me know if there's tooling for local test runs so I can avoid wasting all this compute. Thanks!

Please check the https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-and-writing-tests link I shared before.

Basically:

Run `composer prepare-tests` **once** to set up the MySQL database, and then `composer behat` to run the tests. You can also run only a specific test, see https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-the-test-suite

Alternatively: use `WP_CLI_TEST_DBTYPE=sqlite composer behat` to run tests with SQLite. Then you don't need the `prepare-tests` step.

> Seeing failed tests outside the scope of the PR:

Yes, looks like a newly failing test even on trunk, see https://github.com/wp-cli/extension-command/actions/runs/12247875028/job/34166552233#step:11:91

Please ignore it.

> Looks like Behat is 3.7.0 which doesn't contain diff tooling from the latest versions.

They have new diff features? Do you have a link? Because in https://github.com/wp-cli/wp-cli-tests/issues/17 many years ago we were looking for that. Would be nice to finally fix this.

> ^ Is this a whitespace issue? Is there a technique to understand how this doesn't match the test?

When you are expecting an issue, use `try` instead of `run`

```diff
Scenario: Missing required inputs
-  When I run `wp plugin uninstall`
+  When I try `wp plugin uninstall`
```

features/plugin-uninstall.feature
dkoston: > > Test that if I have the update_plugins transient that any deleted plugins are removed from it
> 
> You can run `wp transient get update_plugins` in the test and verify that the plugin is not part of the output.
> 

Thanks, I made a check for this


> > Test that if uninstall succeeds and delete fails that the user will get the error for that
> 
> You could for example check STDERR/STDOUT and whether the files still exist.

In this instance, I have no way to cause the `delete_plugin()` function to fail:

https://github.com/wp-cli/extension-command/blob/main/src/Plugin_Command.php#L1477

dkoston:
> > There's no actual validation that any plugins have been installed, removed, what logic branches have been exercised, etc.
> > to validate that things are "actually" happening
> 
> You can also use other commands to verify that a plugin is no longer present, e.g. by verifying whether `wp plugin list` includes the plugin, or if the directory no longer exists.

That makes sense. I'll add some additional testing to validate these scenarios using other wp cli commands

> 
> > I didn't see any tooling to run tests locally. Really sorry about using all these compute cycles on GitHub Actions to get these tests passing. Please let me know if there's tooling for local test runs so I can avoid wasting all this compute. Thanks!
> 
> Please check the https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-and-writing-tests link I shared before.
> 
> Basically:
> 
> Run `composer prepare-tests` **once** to set up the MySQL database, and then `composer behat` to run the tests. You can also run only a specific test, see https://make.wordpress.org/cli/handbook/contributions/pull-requests/#running-the-test-suite
> 
> Alternatively: use `WP_CLI_TEST_DBTYPE=sqlite composer behat` to run tests with SQLite. Then you don't need the `prepare-tests` step.
> 

Ah thanks, sorry, I stopped reading at the Behat section. Apologies


> > Seeing failed tests outside the scope of the PR:
> 
> Yes, looks like a newly failing test even on trunk, see https://github.com/wp-cli/extension-command/actions/runs/12247875028/job/34166552233#step:11:91
> 
> Please ignore it.

10-4.


> 
> > Looks like Behat is 3.7.0 which doesn't contain diff tooling from the latest versions.
> 
> They have new diff features? Do you have a link? Because in [wp-cli/wp-cli-tests#17](https://github.com/wp-cli/wp-cli-tests/issues/17) many years ago we were looking for that. Would be nice to finally fix this.

Yea, it was added in https://github.com/Behat/Behat/pull/1532

It looks like there's a PR to introduce newer Behat that's blocked due to php version support: https://github.com/wp-cli/wp-cli-tests/pull/228



> 
> > ^ Is this a whitespace issue? Is there a technique to understand how this doesn't match the test?
> 
> When you are expecting an issue, use `try` instead of `run`
> 
> ```diff
> Scenario: Missing required inputs
> -  When I run `wp plugin uninstall`
> +  When I try `wp plugin uninstall`
> ```

Ah, thanks. Apologies again for being a newbie to all this stuff.

dkoston:
@swissspidy this should be good to go. Thanks for all the help.

I couldn't find a way to populate site transients so I removed that assertion:

```
002 Scenario: Uninstall an installed plugin should uninstall, delete files, and remove from transient update_plugins # features/plugin-uninstall.feature:6
      When I run `wp transient get --network update_plugins`                                                         # features/plugin-uninstall.feature:17
        $ wp transient get --network update_plugins

        Warning: Transient with key "update_plugins" is not set.
        cwd: /var/folders/lm/gzk1q0b95491ctlw16p4_4_m0000gn/T/wp-cli-test-run--6758a46b022fc0.22076735/
        run time: 0.23778200149536
        exit status: 0 (RuntimeException)
```

but otherwise added some checks around more logic branches and file presence

features/plugin-uninstall.feature
swissspidy: Makes sense, I think it's ok not having a test for that last part.

swissspidy:
> Yea, it was added in https://github.com/Behat/Behat/pull/1532

Looks like that's not in a tagged release yet, but maybe we can manually add these changes to our tooling or so. Thanks for pointing it out!

> this should be good to go. Thanks for all the help.

Amazing! Thank **you** for working on this!

> I couldn't find a way to populate site transients

`wp plugin status` should do that. I'll give it a try.

swissspidy (approved):

Issue #428, opened 2024-07-17
Add `requires_php` to the `wp plugin list --update=available` output


dd32:
The following command can be used to list out plugin updates:
```
wp plugin list --update=available
```

but it includes plugins which are incompatible with the current site, due to either the `requires` or `requires_php` field values.

It would be helpful to be able to add the following fields to the output if supported:
 - requires_php
 - requires
 - tested
 - requires_plugins (plugin dependencies)

The update API appears to only return `requires` for no-update plugins and not the `requires_php`/`tested`/`requires_plugins` (this can be fixed) but expanding the data would be helpful.

It would additionally be useful to be able to filter plugins based on those fields, such as `--update=available --canbeinstalled` or something of the sort, for example, if a plugin requires PHP 8 but I'm running PHP 7.4 I don't want it included. I don't know how best to achieve that.

mrsdizzie:
I'd like to work on this, but I'm not sure the premise of this issue is true. Does it actually show available updates where the minimum requirements aren't meant? In my experience, the opposite is true and what I want to address:

WordPress / WP CLI shows there are no updates available, but the reality is the plugin has been updated yet the update can't be installed on the site. I'm not sure the best way to reflect that in the wp cli output.

I'm not sure the best way to add `requires` `requires_php` etc.. as listed above because WordPress only provides those details for the very latest version of the plugin published upstream. It isn't possible to get them for any other version. Also I'm not really sure how this works for plugins that aren't in the wordpress.org repo

I would like to figure out the best way to display to a user that there is a newer version of the plugin available upstream, but it can't be installed because either the version of PHP or WordPress is too low. Somewhat related to #436 but I want `wp plugin list` to do its best possible to let a user know if they are running the latest available version of all plugins currently present on the site (as it used to before requirements were introduced).

So it needs a way to show that there is a newer version available upstream, even if there is no update available for this particular site. I suppose it should also then explain why that newer version is not available. What might that look like?

dd32:
> I'd like to work on this, but I'm not sure the premise of this issue is true. Does it actually show available updates where the minimum requirements aren't meant? In my experience, the opposite is true and what I want to address:

Yes, it does.

Prior to https://github.com/WordPress/wordpress-develop/commit/b00097b75832adebe5682723ea8eb2a9816db55c  incompatible updates (WP version) were not shown, but since then the API has returned incompatible-with-your-WP. I believe the site will show a notice to that effect.

[...]

PR #440, opened 2024-12-24
Support requires and requires_php in plugin/theme list and update command

mrsdizzie:
This introduces a new update sate of 'unavailable' which is when there is a new version of a theme/plugin upstream but the local WordPress / PHP requirements do not meet the requirements set by the authors.

What this changes:

Changes the `update` field of an item from a bool which laters gets turned into 'available' or 'none' to a string which will be either `'available'`, `'none'`, or `'unavailable'`. New `unavailable` is based off a check against WordPress and PHP requirements via the `requires` and `requires_php` fields.

```
$ wp plugin list
+----------------+----------+-------------+---------+----------------+-------------+----------+--------------+
| name           | status   | update      | version | update_version | auto_update | requires | requires_php |
+----------------+----------+-------------+---------+----------------+-------------+----------+--------------+
| akismet        | inactive | available   | 5.1     | 5.3.5          | off         | 5.8      | 5.6.20       |
| edit-flow      | inactive | none        | 0.9.9   |                | off         | 6.0      | 8.0          |
| wp-super-cache | inactive | unavailable | 1.9.4   | 1.12.4         | off         | 6.5      | 7.0          |
+----------------+----------+-------------+---------+----------------+-------------+----------+--------------+
```

```
$ wp plugin update wp-super-cache
Warning: wp-super-cache: This update requires WordPress version 6.5, but the version installed is 6.2.
Error: No plugins updated.
```

This works similarly for PHP versions too:
```
$ php -v
PHP 7.4.33 (cli) (built: Dec 16 2024 20:48:31) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.33, Copyright (c), by Zend Technologies
$ wp plugin list
+----------------+----------+-------------+---------+----------------+-------------+----------+--------------+
| name           | status   | update      | version | update_version | auto_update | requires | requires_php |
+----------------+----------+-------------+---------+----------------+-------------+----------+--------------+
| akismet        | inactive | available   | 5.1     | 5.3.5          | off         | 5.8      | 5.6.20       |
| edit-flow      | inactive | unavailable | 0.9.8   | 0.9.9          | off         | 6.0      | 8.0          |
| wp-super-cache | inactive | unavailable | 1.9.4   | 1.12.4         | off         | 6.5      | 7.0          |
+----------------+----------+-------------+---------+----------------+-------------+----------+--------------+
$ wp plugin update edit-flow
Warning: edit-flow: This update requires PHP version 8.0, but the version installed is 7.4.
Error: No plugins updated.
```

This also fixes issues like #428 by removing incompatible updates from a command like:
```
$ wp plugin list --update=available
+---------+----------+-----------+---------+----------------+-------------+----------+--------------+
| name    | status   | update    | version | update_version | auto_update | requires | requires_php |
+---------+----------+-----------+---------+----------------+-------------+----------+--------------+
| akismet | inactive | available | 5.1     | 5.3.5          | off         | 5.8      | 5.6.20       |
+---------+----------+-----------+---------+----------------+-------------+----------+--------------+
```

This also adds the fields `requires` and `requires_php`. While these aren't default fields,  I did change it so that it will include them by default if one of the plugins has an update sate of unavailable (so it is immediately clear why the update is unavailable). 

I've also written it so that `requires` and `requires_php` fields will use either what is provided for the latest version in the case of an update (unavailable or not) or whatever is listed in the current plugin file.

This means that in the case of 'none' it i showing the requirements for the current version, while in the case of 'available' or 'unavailable' it is showing the requirements for the `update_version`. This could be a bit confusing, but it made even less sense the other way to have it list an update as 'unavailable' while showing the older and already met requirements from the currently installed version. Those were just my initial idea, so any feedback welcome there. Similar to the field names, etc...

Finally, I couldn't find much documentation about the wordpress.org update API, but what I saw in testing is:

For plugins, the wordpress.org api will check the version of WordPress sent in the header of the request from wp core and if the latest upstream version requires a newer version of WordPress, it moves that plugin into the `no_update` section of the response, along with any other plugins that just don't have an update available. It doesn't know the current version of PHP, so it can't check that. The current situation is that if an upstream update requires a newer version of WordPress then WP CLI doesn't see it at all, because it doesn't look for it in the `no_updates` section of the response. If it requires a newer version of PHP, WP CLI will show it as an available update because it doesn't know about the PHP incompatibility.

For themes, it doesn't seem to do that same filtering.

This PR fixes both those cases by correctly showing the updates as `unavailable` rather than not existing or ready to install. My main use case is that I want to be able to run `wp plugin list` and see if a site is running the latest version of any installed plugin or theme. If not, I want to know if it is because the updates haven't been applied or if they are held back because of requirement changes in newer versions.

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -271,6 +271,8 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 				'file'               => $file,
 				'auto_update'        => false,
 				'tested_up_to'       => '',
+				'requires'           => '',
+				'requires_php'       => '',
 				'wporg_status'       => $wporg_info['status'],
 				'wporg_last_updated' => $wporg_info['last_updated'],
 			);
@@ -293,6 +295,8 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 				'auto_update'        => false,
 				'author'             => $item_data['Author'],
 				'tested_up_to'       => '',
+				'requires'           => '',
+				'requires_php'       => '',
 				'wporg_status'       => '',
 				'wporg_last_updated' => '',
 			];
@@ -740,6 +744,8 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	}
 
 	protected function get_item_list() {
+		global $wp_version;
+
 		$items           = [];
 		$duplicate_names = [];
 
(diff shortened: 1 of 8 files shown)


swissspidy:
Thanks a lot for working on this!

I haven't tested it yet, but it looks very promising already. It will be very useful for better managing updates when knowing that newer (albeit incompatible) versions are available.

On that note, what do you think about renaming `unavailable` to `incompatible`?

IMO `unavailable` sounds too similar to `none`, almost a synonym in that context.

src/Plugin_Command.php
```diff
@@ -760,29 +766,62 @@ protected function get_item_list() {
...
+			$requires_php = isset( $update_info ) && isset( $update_info['requires_php'] ) ? $update_info['requires_php'] : null;
+
+			// If an update has requires_php set, check to see if the local version of PHP meets that requirement
+			// The plugins update API already filters out plugins that don't meet WordPress requirements, but does not
+			// filter out plugins based on PHP requirements -- so we must do that here
+			$compatible_php = empty( $requires_php ) || version_compare( PHP_VERSION, $requires_php, '>=' );
+
+			if ( ! $compatible_php ) {
+				$update                    = 'unavailable';
+				$update_unavailable_reason = "Requires a newer version of PHP [$requires_php] than available [$php_version]";
```
swissspidy, line 787: Nit/curious: why square brackets instead of parentheses? Is that a common style elsewhere in the project?

mrsdizzie:
> On that note, what do you think about renaming `unavailable` to `incompatible`?
> 
> IMO `unavailable` sounds too similar to `none`, almost a synonym in that context.

Interesting, I picked `unavailable` because its an antonym of the other option `available` but I can see that : ) 

`incompatible` could work too and I don't really have a strong preference on that. I think whatever word also gives the most confidence that wp cli isn't going to install it if you do something like `wp plugin update --all`

src/Plugin_Command.php
mrsdizzie: This is just a personal practice hold over from writing lots of code that goes into log files where square brackets are often used over parentheses. It probably doesn't need either of them here though, will change

mrsdizzie: Does this seem OK:
```
Warning: wp-super-cache: This update requires WordPress version 6.5, but the version installed is 6.2.
```
(Changing the text means changing all the test, so we can figure out a good string here first

swissspidy: Looks good to me!

swissspidy:
> > On that note, what do you think about renaming `unavailable` to `incompatible`?
> > IMO `unavailable` sounds too similar to `none`, almost a synonym in that context.
> 
> Interesting, I picked `unavailable` because its an antonym of the other option `available` but I can see that : )
> 
> `incompatible` could work too and I don't really have a strong preference on that. I think whatever word also gives the most confidence that wp cli isn't going to install it if you do something like `wp plugin update --all`

@wp-cli/committers @wp-cli/maintainers any preference here on the wording? Also generally a review here is appreciated :)

mrsdizzie:
@swissspidy I updated tests for this to use new HTTP mock feature which works well here (and makes testing more reliable). The only remaining PHP 5.6 / PHP 3.7 issues are because we use [Moina](https://wordpress.org/themes/moina/) in many tests and it does not support WP 3.7 so it now correctly throws an error. If there is an interest in merging this, I'm happy to go through and fix the tests before merge (probably just replacing with twentyten).

swissspidy:
I'm definitely interested in merging this!

Replacing the theme with another one or just adding some `@requires-wp..` tag would both be fine IMO.

As for the unavailable vs incompatible question, I don't feel strongly. Let's stick with unavailable for now.

mrsdizzie:
I added them behind require flags because trunk actually includes *every* previous default theme (unlike releases) and gives different output when a theme is already installed. All tests passing now 🚥

swissspidy (approved):
Amazing, thanks for this!

PR #443, opened 2025-03-10
Fix failing tests

swissspidy:
Fixes some tests on WordPress 4.9.

There's also one failing test against WordPress trunk, but not sure why.

Started on the 5th but was fine on the 4th. See https://github.com/wp-cli/extension-command/actions/runs/13644528411
and https://github.com/wp-cli/extension-command/actions/runs/13666785798.

Thought maybe https://core.trac.wordpress.org/changeset/59926 is related. Otherwise couldn't find a suspicious commit.

https://github.com/wp-cli/extension-command/actions/runs/13755011683/job/38461015602#step:11:127

### DIFF
--- a/features/plugin-update.feature
+++ b/features/plugin-update.feature
@@ -95,7 +95,8 @@ Feature: Update WordPress plugins
       0.5.2
       """
 
-  @require-wp-4.0
+  # Akismet currently requires WordPress 5.8
+  @require-wp-5.8
   Scenario: Update a plugin to its latest minor release
     Given a WP install
     And I run `wp plugin install --force akismet --version=2.5.4`
@@ -196,7 +197,8 @@ Feature: Update WordPress plugins
       Error: Can't find the requested plugin's version 2.5.4 in the WordPress.org plugin repository (HTTP code 404).
       """
 
-  @require-wp-4.7
+  # Akismet currently requires WordPress 5.8
+  @require-wp-5.8
   Scenario: Plugin updates that error should not report a success
     Given a WP install
     And I run `wp plugin install --force akismet --version=4.0`
@@ -223,6 +225,8 @@ Feature: Update WordPress plugins
       Success:
       """
 
+  # Akismet currently requires WordPress 5.8, so there's a warning because of it.
+  @require-wp-5.8
   Scenario: Excluding a missing plugin should not throw an error
     Given a WP install
     And I run `wp plugin update --all --exclude=missing-plugin`
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -107,7 +107,8 @@ Feature: Manage WordPress plugins
     And STDOUT should be empty
     And the return code should be 1
 
-  @require-wp-4.0
+  # WordPress Importer currently requires at least WP 5.2.
+  @require-wp-5.2
   Scenario: Install a plugin, activate, then force install an older version of the plugin
     Given a WP install
 
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -579,7 +579,8 @@ Feature: Manage WordPress plugins
       """
     And the return code should be 0
 
-  @require-wp-4.7
+  # Akismet currently requires WordPress 5.8, so there's a warning because of it.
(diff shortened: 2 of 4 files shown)


swissspidy:
Here's the relevant code for that trunk-related test:

https://github.com/wp-cli/extension-command/blob/33afe4f86812a3173c20a584c6a070e68dfafc4c/src/Plugin_Command.php#L1250-L1261

mrsdizzie:
I was able to set up php 7.2 locally and I can't reproduce any errors when running a copy of WordPress 4.9.26 but when trying to run some of the tests I am seeing these failures:
```
Notice: An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="https://wordpress.org/support/">support forums</a>. (WordPress could not establish a secure connection to WordPress.org. Please contact your server administrator.) in /private/var/folders/v6/gqkd9_rd74171hsndwtjrxww0000gn/T/wp-cli-test-run--67cf0b401e37f3.85700623/wp-includes/update.php on line 347
```
It works when running the command against a local WordPress install outside of the testing environment so it isn't a general problem...

swissspidy:
Sure yeah that's possible.

We could add a change to https://github.com/wp-cli/.github that runs test with the `--format=pretty` argument when re-running tests with debug logging, see this screenshot:

Screenshot 2025-03-10 at 17 21 00

Related, there's also this long-standing bug to have better output when comparing actual output vs expected output, see https://github.com/wp-cli/wp-cli-tests/issues/17

mrsdizzie:
oo neat! -- better output would indeed be nice but at least showing the specific error/stack trace is helpful for seeing why something failed.

swissspidy:
https://github.com/wp-cli/.github/pull/122

mrsdizzie:
I think my issue is local to my php 7.2 setup somehow, so maybe a red herring. Looking at failures this is probably similarly related to plugins that have since added higher WordPress requirements than 4.9 and generate different output on 4.9 now. I'll try and get php 7.2 working right so I can run the tests locally

swissspidy:
I thought so too but e.g. [this result](https://github.com/wp-cli/extension-command/actions/runs/13755011683/job/38461011029#step:11:86) is confusing.

```
004 Scenario: Create, activate and check plugin status # features/plugin.feature:3
      Then STDOUT should be a table containing rows:   # features/plugin.feature:65
        name	status	update	version	update_version	auto_update	requires	requires_php
        akismet	inactive	unavailable	4.0.1	5.3.7	off	5.8	5.6.20
        hello	inactive	available	1.6	1.7.2	off	4.6	
        Zombieland	active	none	0.1.0		off		
        no-mail	must-use				off		
        polyfills	must-use				off		
         (Exception)
```

And here's that line in the test:

https://github.com/wp-cli/extension-command/blob/33afe4f86812a3173c20a584c6a070e68dfafc4c/features/plugin.feature#L64-L67

It only tests the line for a plugin that's not in the repo.

But if it's really just about incompatible version we can bump the minimum requirement of those affected tests.

In the meantime, running the tests here with debug logging enabled:

https://github.com/wp-cli/extension-command/actions/runs/13764647737/job/38509224929?pr=443

The extra output isn't really helpful though 🤔 

(Note to self: this apparently combines `--format progress` with `--format=pretty` so there are some dots here and there 🤦 )

mrsdizzie:
Maybe since the test is missing the new columns it is failing

swissspidy:
Then it should fail on other versions too, no?

Feels like I really need to try to install 7.2 again 🙃

swissspidy:
> So for this particular test, if the goal is to really only check the status of one plugin it should maybe be using `wp plugin status Zombieland` because the table columns of `wp plugin list` are now variable depending on what version of WordPress is used since only an older one is going to trigger the new columns to be displayed.

Makes sense to me 👍

mrsdizzie (approved):
@swissspidy all better ✅

swissspidy:
@mrsdizzie oh neat, thanks a lot!

PR #446, opened 2025-03-19
Don't use transient values when doing wp plugin|theme list

mrsdizzie:
This changes the default behavior to always check wordpress.org for the latest information when doing wp plugin|theme list unless the existing `--skip-update-check` flag is passed.

This came out of the discussion in #426 and replaces that PR. The main idea is that it doesn't make any sense to have both `--force-check` and `--skip-update-check` flags while also having the default behavior be unpredictable depending on the state of the transient.

I created a new PR since pretty much all of the code is test code and that is different now that we are changing the default behavior instead of adding a flag.

### DIFF
--- a/features/plugin.feature
+++ b/features/plugin.feature
@@ -318,6 +318,27 @@ Feature: Manage WordPress plugins
       Automattic
       """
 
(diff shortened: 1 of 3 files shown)


swissspidy (approved):
Nice one! 🚀

PR #453, opened 2025-05-18
PHPStan level 9

swissspidy:
Mostly doc block improvements, but also fixes a regression from #451

### DIFF
--- a/src/Plugin_Command.php
+++ b/src/Plugin_Command.php
@@ -1,5 +1,6 @@
 <?php
 
+use WP_CLI\CommandWithUpgrade;
 use WP_CLI\ParsePluginNameInput;
 use WP_CLI\Utils;
 use WP_CLI\WpOrgApi;
@@ -43,9 +44,9 @@ use function WP_CLI\Utils\normalize_path;
  * @package wp-cli
  *
  * @phpstan-type PluginInformation object{name: string, slug: non-empty-string, version: string, new_version: string, download_link: string, requires_php?: string, requires?: string, package: string}&\stdClass
+ * @extends CommandWithUpgrade<string,>
  */
-class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
-
+class Plugin_Command extends CommandWithUpgrade {
 	use ParsePluginNameInput;
 
 	protected $item_type         = 'plugin';
@@ -72,6 +73,8 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 * Plugin fetcher instance.
 	 *
 	 * @var \WP_CLI\Fetchers\Plugin
+	 *
+	 * @phpstan-ignore property.phpDocType (To be fixed with in https://github.com/wp-cli/wp-cli/pull/6096)
 	 */
 	protected $fetcher;
 
@@ -354,9 +357,9 @@ class Plugin_Command extends \WP_CLI\CommandWithUpgrade {
 	 * @param array $args
 	 * @param array $assoc_args
 	 */
-	public function activate( $args, $assoc_args = array() ) {
-		$network_wide = (bool) Utils\get_flag_value( $assoc_args, 'network', false );
-		$all          = (bool) Utils\get_flag_value( $assoc_args, 'all', false );
+	public function activate( $args, $assoc_args = [] ) {
+		$network_wide = Utils\get_flag_value( $assoc_args, 'network', false );
+		$all          = Utils\get_flag_value( $assoc_args, 'all', false );
 		$all_exclude  = Utils\get_flag_value( $assoc_args, 'exclude', '' );
 
 		/**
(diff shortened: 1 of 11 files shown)


src/Plugin_Command.php
```diff
@@ -875,7 +904,7 @@ protected function get_item_list() {
 
 					if ( isset( $plugin_update_info->requires ) && version_compare( $wp_version, $requires, '>=' ) ) {
 						$reason = "This update requires WordPress version $plugin_update_info->requires, but the version installed is $wp_version.";
-					} elseif ( ! isset( $update_info['package'] ) ) {
+					} elseif ( ! isset( $plugin_update_info->package ) ) {
```
swissspidy, line 907: `$update_info` does not exist at this point.

Introduced in #451

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -197,13 +203,16 @@ public function install( $args, $assoc_args ) {
 
 				$filter = false;
 				// If a GitHub URL, do some guessing as to the correct plugin/theme directory.
-				if ( $is_remote && 'github.com' === $this->parse_url_host_component( $slug, PHP_URL_HOST )
+				if ( $is_remote && 'github.com' === \WP_CLI\Utils\parse_url( $slug, PHP_URL_HOST )
```
schlessera, line 206:
```suggestion
				if ( $is_remote && 'github.com' === Utils\parse_url( $slug, PHP_URL_HOST )
```

src/Plugin_Command.php
```diff
@@ -1233,7 +1273,11 @@ public function uninstall( $args, $assoc_args = array() ) {
 
 					$json_translation_files = glob( WP_LANG_DIR . '/plugins/' . $plugin_slug . '-' . $translation . '-*.json' );
 					if ( $json_translation_files ) {
-						array_map( array( $wp_filesystem, 'delete' ), $json_translation_files );
+						/**
+						 * @var callable $callback
+						 */
+						$callback = array( $wp_filesystem, 'delete' );
```
schlessera, line 1279:
```suggestion
						$callback = [ $wp_filesystem, 'delete' ];
```

src/Theme_Command.php
```diff
@@ -211,8 +213,11 @@ protected function get_all_items() {
 	 *
 	 *     $ wp theme activate twentysixteen
 	 *     Success: Switched to 'Twenty Sixteen' theme.
+	 *
+	 * @param array $args
+	 * @param array $assoc_args
 	 */
-	public function activate( $args = array() ) {
+	public function activate( $args, $assoc_args = [] ) {
```
schlessera, line 220: This changed `$args` from optional to required. Was this intended?

src/Theme_Mod_Command.php
```diff
@@ -232,6 +232,9 @@ public function remove( $args = array(), $assoc_args = array() ) {
 	 *     # Set theme mod
 	 *     $ wp theme mod set background_color 000000
 	 *     Success: Theme mod background_color set to 000000.
+	 *
+	 * @param string[] $args
```
schlessera, line 236: Different types then for `activate()` above. We should probably define and stick to a consistent type hint for command methods.

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -80,6 +84,8 @@ abstract protected function status_single( $args );
 
 	abstract protected function install_from_repo( $slug, $assoc_args );
 
+	abstract public function activate( $args, $assoc_args = [] );
```
schlessera, line 87: Should probably stick to consistent command signature as well.

src/WP_CLI/CommandWithUpgrade.php
```diff
@@ -197,13 +203,16 @@ public function install( $args, $assoc_args ) {
...
 						// Don't attempt to rename ZIPs uploaded to the releases page or coming from a raw source.
 						&& ! preg_match( '#github\.com/[^/]+/[^/]+/(?:releases/download|raw)/#', $slug ) ) {
 
 					$filter = function ( $source ) use ( $slug ) {
-
-						$slug_dir = Utils\basename( $this->parse_url_host_component( $slug, PHP_URL_PATH ), '.zip' );
+						/**
+						 * @var string $path
+						 */
+						$path     = \WP_CLI\Utils\parse_url( $slug, PHP_URL_PATH );
```
schlessera, line 214:
```suggestion
						$path     = Utils\parse_url( $slug, PHP_URL_PATH );
```

schlessera (changes requested):

src/Theme_Command.php
swissspidy: Yes, because it is actually required as per the synopsys, and WP-CLI always passes both arguments to commands anyway. I've never seen `$args` or `$assoc_args` be null actually, so should even change this to `( $args, $assoc_args )`.

src/Theme_Mod_Command.php
swissspidy: Done!

src/WP_CLI/CommandWithUpgrade.php
swissspidy: Done!

swissspidy: So `$assoc_args` is sometimes optional because the class is extended. The method is sometimes called with 1 parameter from within the sub class itself.

src/Plugin_Command.php
mrsdizzie: 🤦‍♀️ yes that is a typo since these are similarly named. I should set up PHPStan (!)

PR #456, opened 2025-06-05
Fix output formats for the `theme mod get` command

### LINKED ISSUE #96: wp theme mod get --all produces malformed output for non-tabular formats
The `wp theme mod get --all` command formats its output in a clever way (this is the first time I've seen this particular formatting, I'm not sure if it's used elsewhere) by listing the first level keys of a serialised option value as its own element.

Example:

```
+-------------------------+--------------------------------------------------------------------------------------------------+
| key                     | value                                                                                            |
+-------------------------+--------------------------------------------------------------------------------------------------+
| 0                       |                                                                                                  |
| background_color        | ffffff                                                                                           |
| nav_menu_locations      | =>                                                                                               |
|     primary             | 39                                                                                               |
| color_scheme            | yellow                                                                                           |
| sidebar_textcolor       | #111111                                                                                          |
| header_background_color | #ffffff                                                                                          |
| background_image        |                                                                                                  |
| background_attachment   | scroll                                                                                           |
| sidebars_widgets        | =>                                                                                               |
|     time                | 1458419959                                                                                       |
|     data                | {"wp_inactive_widgets":["calendar-3","search-3","search-5","categories-2","recent-comments-3","p |
|                         | ages-2","calendar-2","archives-2","meta-2","search-2","text-2","recent-posts-2","recent-comments |
|                         | -2","rss-2","tag_cloud-2","nav_menu-2","nav_menu-4"],"sidebar-1":["search-7","tag_cloud-5"],"sid |
|                         | ebar-2":["search-7","tag_cloud-5"],"sidebar-3":null}                                             |
| custom_css_post_id      | -1                                                                                               |
+-------------------------+--------------------------------------------------------------------------------------------------+
```

Observer that the `primary` element from the `nav_menu_locations` option, and the `time` and `data` elements from the `sidebars_widgets` option are displayed as their own sub-elements in the table.

This output works okay for the tabular output, but for other formats it's broken.

Example of broken CSV output:

```csv
key,value
0,
background_color,ffffff
nav_menu_locations,=>
"    primary",39
color_scheme,yellow
sidebar_textcolor,#111111
header_background_color,#ffffff
background_image,
background_attachment,scroll
sidebars_widgets,=>
"    time",1458419959
"    data","{""wp_inactive_widgets"":[""calendar-3"",""search-3"",""search-5"",""categories-2"",""recent-comments-3"",""pages-2"",""calendar-2"",""archives-2"",""meta-2"",""search-2"",""text-2"",""recent-posts-2"",""recent-comments-2"",""rss-2"",""tag_cloud-2"",""nav_menu-2"",""nav_menu-4""],""sidebar-1"":[""search-7"",""tag_cloud-5""],""sidebar-2"":[""search-7"",""tag_cloud-5""],""sidebar-3"":null}"
custom_css_post_id,-1
```

philipjohn:
Fixes #96 

I maintained but improved the space-indented formatting for the table format as folks might be using this (as per @schlessera's comment) and then fixed the other formats by using dot notation to represent each field, given the formatter needs arrays with `key` and `value`.

For the table format, I opted to remove the `=>` for the value  to represent an array. I felt it was harder to read than simply leaving the value blank and letting the indent do the visual work.

For all formats booleans and empty arrays are represented by a string (e.g. `[empty array]`. There may be a better way to represent these, and I did consider `var_export()` but others may have better ideas.

When specific mods are requested but don't exist, they are included in the output with `null` as the value except for CSV output which just has an empty string.

New Behat tests are included for each of the formats (table, csv, json, yaml).

Here are some examples of the new output:

```
$ wp eval 'print_r(get_theme_mods());'
Array
(
    [0] => 
    [nav_menu_locations] => Array
        (
        )

    [sidebars_widgets] => Array
        (
            [time] => 1749114163
            [data] => Array
                (
                    [wp_inactive_widgets] => Array
                        (
                            [0] => block-2
                            [1] => block-3
                            [2] => block-4
                            [3] => block-5
                            [4] => block-6
                        )

                    [sidebar-1] => Array
                        (
                            [0] => block-7
                        )

                )

        )

)
```

```
$ wp theme mod get --all --format=table
+-------------------------+---------------+
| key                     | value         |
+-------------------------+---------------+
| 0                       | [false]       |
| nav_menu_locations      | [empty array] |
| sidebars_widgets        |               |
|   time                  | 1749114163    |
|   data                  |               |
|     wp_inactive_widgets |               |
|       0                 | block-2       |
|       1                 | block-3       |
|       2                 | block-4       |
|       3                 | block-5       |
|       4                 | block-6       |
|     sidebar-1           |               |
|       0                 | block-7       |
+-------------------------+---------------+
```

```
$ wp theme mod get --all --format=csv
key,value
0,[false]
nav_menu_locations,"[empty array]"
sidebars_widgets.time,1749114163
sidebars_widgets.data.wp_inactive_widgets.0,block-2
sidebars_widgets.data.wp_inactive_widgets.1,block-3
sidebars_widgets.data.wp_inactive_widgets.2,block-4
sidebars_widgets.data.wp_inactive_widgets.3,block-5
sidebars_widgets.data.wp_inactive_widgets.4,block-6
sidebars_widgets.data.sidebar-1.0,block-7
```

```
$ wp theme mod get --all --format=yaml
---
- 
  key: "0"
  value: '[false]'
- 
  key: nav_menu_locations
  value: '[empty array]'
- 
  key: sidebars_widgets.time
  value: 1749114163
- 
  key: sidebars_widgets.data.wp_inactive_widgets.0
  value: block-2
- 
  key: sidebars_widgets.data.wp_inactive_widgets.1
  value: block-3
- 
  key: sidebars_widgets.data.wp_inactive_widgets.2
  value: block-4
- 
  key: sidebars_widgets.data.wp_inactive_widgets.3
  value: block-5
- 
  key: sidebars_widgets.data.wp_inactive_widgets.4
  value: block-6
- 
  key: sidebars_widgets.data.sidebar-1.0
  value: block-7
```

```
$ wp theme mod get --all --format=json | jq .
[
  {
    "key": "0",
    "value": "[false]"
  },
  {
    "key": "nav_menu_locations",
    "value": "[empty array]"
  },
  {
    "key": "sidebars_widgets.time",
    "value": 1749114163
  },
  {
    "key": "sidebars_widgets.data.wp_inactive_widgets.0",
    "value": "block-2"
  },
  {
    "key": "sidebars_widgets.data.wp_inactive_widgets.1",
    "value": "block-3"
  },
  {
    "key": "sidebars_widgets.data.wp_inactive_widgets.2",
    "value": "block-4"
  },
  {
    "key": "sidebars_widgets.data.wp_inactive_widgets.3",
    "value": "block-5"
  },
  {
    "key": "sidebars_widgets.data.wp_inactive_widgets.4",
    "value": "block-6"
  },
  {
    "key": "sidebars_widgets.data.sidebar-1.0",
    "value": "block-7"
  }
]
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced `wp theme mod get` output across table, CSV, JSON, and YAML, including dot-notated handling of nested theme mods.
  * Non-table formats now preserve native value types (including booleans and empty arrays) for machine-readable results.
* **Bug Fixes**
  * Standardized how unset or missing theme mods are represented across formats (e.g., empty cells vs `null` vs YAML `value: null`).
* **Tests**
  * Expanded coverage to validate empty and populated states, `--all`, and single-mod retrieval with `--field=`, including correct format-specific serialization.


src/Theme_Mod_Command.php
```diff
@@ -85,49 +85,113 @@ public function get( $args = array(), $assoc_args = array() ) {
...
-		// For unset mods, show blank value.
-		foreach ( $args as $mod ) {
-			if ( ! isset( $mods[ $mod ] ) ) {
-				$list[] = [
-					'key'   => $mod,
-					'value' => '',
-				];
-			}
+		// Take our Formatter-friendly list and adjust it according to the requested format.
+		switch ( \WP_CLI\Utils\get_flag_value( $assoc_args, 'format' ) ) {
```
schlessera, line 108: This should be simplified by importing `WP_CLI\Utils` and then using `Utils\get_flag_value()` here instead.

src/Theme_Mod_Command.php
```diff
@@ -85,49 +85,113 @@ public function get( $args = array(), $assoc_args = array() ) {
...
+			// For tables we use a double space to indent child items.
+			case 'table':
+				$mod_list = array_map(
+					function ( $item ) {
```
schlessera, line 112: This should be a `static` function.

schlessera (dismissed):
Looks good, only minor nitpicks for the code. Also, it would be good to have at least 1 feature test covering this.

src/Theme_Mod_Command.php
```diff
@@ -85,49 +85,113 @@ public function get( $args = array(), $assoc_args = array() ) {
...
-		$list = array();
-		$mods = get_theme_mods();
+		// This array will hold the list of theme mods in a format suitable for the WP CLI Formatter.
+		$mod_list = array();
+
+		// If specific mods are requested, filter out any that aren't requested.
+		$mods = ! empty( $args ) ? array_intersect_key( get_theme_mods(), array_flip( $args ) ) : get_theme_mods();
 		if ( ! is_array( $mods ) ) {
-			// If no mods are set (perhaps new theme), make sure foreach still works.
+			// If no mods are set (perhaps new theme), make sure array_intersect_key still works.
```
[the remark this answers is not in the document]
philipjohn: Turns out we don't even need to cast here - `get_theme_mods()` already does the necessary work for us.

swissspidy:
This is lacking tests & is breaking existing ones. Will see if I can pick this up to bring it over the finish line.

schlessera (commented):
Thanks @philipjohn, this is a solid cleanup and the per-format Behat coverage is exactly what #96 needed.

One blocker before this can go in: the WP 4.9 / PHP 7.2 job is failing, and it's a real regression rather than a flake. The `--all` branch now does:

```php
} else {
    $mods = get_theme_mods();
}
```

`get_theme_mods()` returns `false` (not an array) when a site has no mods set, which is the case on a fresh install with older WP. The following `foreach ( $mods as ... )` then warns with "Invalid argument supplied for foreach()", that lands on STDERR, and the very first `wp theme mod get --all` step fails. Newer WP ships default mods so the array path is taken there, which is why only the 4.9 matrix trips. The previous code guarded against exactly this and the refactor dropped the guard. Restoring it fixes the job:

```php
} else {
    $mods = get_theme_mods();
    if ( ! is_array( $mods ) ) {
        $mods = [];
    }
}
```

A couple of smaller things:

- `mod_to_string()` uses long `array()` syntax, while the rest of the file (and our packages in general) uses short `[]`.
- Representing booleans and empty arrays as strings (`[false]`, `[empty array]`) makes sense for the table format, but for json/yaml it throws away the type: a consumer gets the string `"[false]"` instead of `false`. I think we should use native values for the machine-readable formats and keep the string markers for table only.

Once the guard is back and the array syntax is tidied up, we should be good to merge. I'll take the liberty to make the final edits so we can get this included in the next release.

philipjohn:
Ah I never got around to fixing those after WCEU, apologies, and thanks for sorting them out @schlessera !

schlessera:
Thanks for the PR, @philipjohn !

PR #467, opened 2025-10-05
Mock HTTP requests in wporg status tests

### LINKED ISSUE #462: Fix flakey `wporg_status` / `wporg_last_updated` tests
These rely on requests to the WordPress.org API and often fail because those requests fail.

We should use our new HTTP request mocking functionality to make the tests more robust.

swissspidy:
Aims to improve test stability.

Fixes #462

### DIFF
--- a/features/plugin-list-wporg-status.feature
+++ b/features/plugin-list-wporg-status.feature
@@ -13,10 +13,61 @@ Feature: Check the status of plugins on WordPress.org
        * Version:     2.0.2
        */
       """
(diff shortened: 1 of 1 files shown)


features/plugin-list-wporg-status.feature
```diff
@@ -28,13 +79,13 @@ Feature: Check the status of plugins on WordPress.org
     When I run `wp plugin list --fields=name,wporg_last_updated`
     Then STDOUT should be a table containing rows:
       | name                   | wporg_last_updated |
-      | wordpress-importer     | {COMMIT_DATE}      |
+      | wordpress-importer     | 2025-09-26      |
```
mrsdizzie, line 82: Wouldn't this break next time it is updated?

mrsdizzie: oh nvm I now realize the whole point of this : )

PR #474, opened 2025-11-06
Fix `url` field in plugin search

### LINKED ISSUE #470: Documented `url` field cannot be use in `wp plugin search` command's `--fields` option
## Bug Report

According to the latest [documentation](https://developer.wordpress.org/cli/commands/plugin/search/) and `wp help plugin search`, the `--fields` option support the field `url`, which should be the "Plugin's URL on wordpress.org".

In reality, if I add "url" to the `--fields` option, it will always show "Error: Invalid field: url.":
Image

Same can be show for different plugin search
```
root@ae8caccf12cf:/var/www/html/wp-content/plugins# wp plugin search --fields=name,slug,url jetpack
Success: Showing 10 of 703 plugins.
Error: Invalid field: url.
```

Run a plugin search command like:
```bash
wp plugin search --fields=name,slug,url jetpack
```

A list of plugin with all the specified fields (include "url") to appear.

```
OS:	Linux 6.17.5-300.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Oct 23 15:35:13 UTC 2025 x86_64
Shell:	
PHP binary:	/usr/local/bin/php
PHP version:	8.1.33
php.ini used:	
MySQL binary:	/usr/bin/mariadb
MySQL version:	mariadb from 11.8.3-MariaDB, client 15.2 for debian-linux-gnu (x86_64) using  EditLine wrapper
SQL modes:	
WP-CLI root dir:	phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	phar:///usr/local/bin/wp-cli
WP-CLI packages dir:	
WP-CLI cache dir:	/root/.wp-cli/cache
WP-CLI global config:	/root/.wp-cli/config.yml
WP-CLI project config:	
WP-CLI version:	2.12.0
```

Haven't dig into the code myself. Sorry.

Website as of now:
Image

Display of `wp help plugin search`:
Image

Testing search command shows unexpected error:
Image

swissspidy:
Fixes #470

### DIFF
--- a/src/WP_CLI/CommandWithUpgrade.php
+++ b/src/WP_CLI/CommandWithUpgrade.php
@@ -853,9 +853,12 @@ abstract class CommandWithUpgrade extends \WP_CLI_Command {
 		$items = $api->$plural;
 
 		// Add `url` for plugin or theme on wordpress.org.
+		// In older WP versions these used to be objects.
 		foreach ( $items as $index => $item_object ) {
 			if ( is_array( $item_object ) ) {
 				$items[ $index ]['url'] = "https://wordpress.org/{$plural}/{$item_object['slug']}/";
+			} elseif ( $item_object instanceof \stdClass ) {
+				$item_object->url = "https://wordpress.org/{$plural}/{$item_object->slug}/";
 			}
 		}
 


mrsdizzie:
I think this is maybe meant to be manually created based on the slug but gives an error before that point. I haven't gone through the blame, but presumably the error code above this was added later on and why we have this field that doesn't work:

https://github.com/wp-cli/extension-command/blob/67129cad9a956ddad3b99b3dcb9d6882b15b9ef9/src/WP_CLI/CommandWithUpgrade.php?plain=1#L855-L860

Or that it uses so much abstraction that wasn't obvious when something was changed.

mrsdizzie (approved):

PR #528, opened 2026-06-04
Fix change --path parameter to --target-path parameter to avoid clashing with global parameter

Rvouill:
Adds to https://github.com/wp-cli/extension-command/pull/527

https://make.wordpress.org/cli/handbook/references/config/


schlessera (approved):

swissspidy:
There's also another way to to make this argument name work, see https://github.com/wp-cli/wp-cli/pull/6174

